PR #17, opened 2024-12-02
FSPT-141 Tidying up lift and shift of assessment store

srh-sloan:
## Change description
Tidying up the lift and shift of `assessment-store` into `pre-award-stores`:
- deleting files that are replaced by ones at the app top level
- Fixing the unit tests
- Creating migrations for the assessment-store tables


MarcUsher (approved):
YES

PR #25, opened 2024-12-04
Set a valid CPU+Memory combination

samuelhwilliams:
ECS doesn't allow 1024 CPU with 1024 memory. If you want 1024 CPU, the minimum memory allowed is 2048. So let's bump up to that.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html


MarcUsher (approved):

sfount (approved):
The "Memory value" column in the table doesn't quite get this across!

PR #26, opened 2024-12-04
Remove bootstrapping openapi spec/routes

samuelhwilliams:
We added this to test that the new application was functional. Now that we've imported the other three pre-award stores, we don't need this any more.


MarcUsher (approved):
Ah yes thank you! Good to rip this out now!

srh-sloan (approved):

PR #39, opened 2024-12-11
updating script to check is ggis available otherwise add none

nuwan-samarasinghe:
### Change description
This is a hot fix done for the GGIS field which is going to use in future with cabinet office integration (No specific timeframe for this integration) and this field is a non mandatory field so if fund has data it will be saved on the DB and if not it will be a null field. (I think this field came with the data standard changes.)

No unit test needed because there are some unit testes that covering up this field

### How to test
01. Run the fund config all load script

image

### DIFF
--- a/fund_store/db/queries.py
+++ b/fund_store/db/queries.py
@@ -277,7 +277,7 @@ def insert_fund_data(fund_config, commit: bool = True):
         "owner_organisation_name": fund_config["owner_organisation_name"],
         "owner_organisation_shortname": fund_config["owner_organisation_shortname"],
         "owner_organisation_logo_uri": fund_config["owner_organisation_logo_uri"],
-        "ggis_scheme_reference_number": fund_config["ggis_scheme_reference_number"],
+        "ggis_scheme_reference_number": fund_config["ggis_scheme_reference_number"] if 'ggis_scheme_reference_number' in fund_config else None,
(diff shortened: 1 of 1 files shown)


fund_store/db/queries.py
```diff
@@ -277,7 +277,7 @@ def insert_fund_data(fund_config, commit: bool = True):
         "owner_organisation_name": fund_config["owner_organisation_name"],
         "owner_organisation_shortname": fund_config["owner_organisation_shortname"],
         "owner_organisation_logo_uri": fund_config["owner_organisation_logo_uri"],
-        "ggis_scheme_reference_number": fund_config["ggis_scheme_reference_number"],
+        "ggis_scheme_reference_number": fund_config["ggis_scheme_reference_number"] if 'ggis_scheme_reference_number' in fund_config else None,
```
samuelhwilliams, line 280:
```suggestion
        "ggis_scheme_reference_number": fund_config.get("ggis_scheme_reference_number"),
```

maybe this alternative is a bit more concise/pythonic, but both should be fine 👍

samuelhwilliams (approved):

hamzabinkhalid (approved):

fund_store/db/queries.py
nuwan-samarasinghe: @albertkol @samuelhwilliams merge which is good :D

PR #44, opened 2024-12-12
Fspt 177 apply submit to assess

srh-sloan:
### Change description
https://mhclgdigital.atlassian.net/browse/FSPT-177
This PR stops application-store from writing to the import queue for assessment. Instead, it calls assessment-store directly to copy the data into the assessment tables. Covered in this PR:
- Updated `submit_application` to call assessment store functions directly to derive application values and insert the row.
- Moved the `send_submit_notification` code into a separate function for easier readability
- Unit tests for the new submit route and functions
- Updated error handling code so that:
    - Submit will not fail for the user if the notification fails to send, or if we can't derive application values

Out of scope:
- Removing the code to read from the queue in assessment. This de-risks the change when it goes into prod and means no data will be lost if someone submits using the old method as we deploy the change.

### How to test
Pull this branch.
Create and submit an application
You should see it in assessment as normal.
No changes should be visible to the user

N/A

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,4 @@
 {
-    "python.testing.pytestArgs": [
-        "tests"
-    ],
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true,
     "editor.quickSuggestions": {
         "other": true,
         "comments": false,
@@ -16,5 +11,10 @@
       },
       "editor.formatOnSave": true,
       "editor.defaultFormatter": "charliermarsh.ruff"
-    }
+    },
+    "python.testing.pytestArgs": [
+      "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,9 @@
 {
+  "python.testing.pytestArgs": [
+    "tests"
+  ],
+  "python.testing.unittestEnabled": false,
+  "python.testing.pytestEnabled": true,
     "editor.quickSuggestions": {
         "other": true,
         "comments": false,
@@ -11,10 +16,5 @@
       },
       "editor.formatOnSave": true,
       "editor.defaultFormatter": "charliermarsh.ruff"
-    },
-    "python.testing.pytestArgs": [
-      "tests"
-    ],
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    }
 }
--- a/application_store/_helpers/application.py
+++ b/application_store/_helpers/application.py
@@ -74,6 +74,7 @@ def send_submit_notification(
                     "Unknown eoi_decision [{eoi_decision}], unable to send submit notification",
                     extra=dict(eoi_decision=eoi_decision),
                 )
+                return
     else:
         notify_template = Config.NOTIFY_TEMPLATE_SUBMIT_APPLICATION
 
--- a/application_store/_helpers/application.py
+++ b/application_store/_helpers/application.py
@@ -44,38 +44,34 @@ def send_submit_notification(
     round_data,
 ):
     contents = {
-        NotifyConstants.APPLICATION_FIELD: application_with_form_json_and_fund_name,
+        NotifyConstants.APPLICATION_FIELD: create_qa_base64file(application_with_form_json_and_fund_name, True),
         NotifyConstants.MAGIC_LINK_CONTACT_HELP_EMAIL_FIELD: round_data.contact_email,
     }
+    del contents[NotifyConstants.APPLICATION_FIELD]["forms"]
+    full_name = account.full_name
 
     if eoi_results:
         eoi_decision = eoi_results["decision"]
-        if eoi_decision == Decision.FAIL:
-            return  # we don't send emails for a failure
-
-        full_name = (
-            account.full_name
-            if account.full_name
-            else map_application_key_fields(
+        if not full_name:
+            full_name = map_application_key_fields(
                 application_with_form_json,
                 ROUND_ID_TO_KEY_REPORT_MAPPING[application.round_id],
                 application.round_id,
             ).get("lead_contact_name", "")
-        )
 
-        if Decision(eoi_decision) == Decision.PASS:  # EOI Full pass
-            notify_template = Config.NOTIFY_TEMPLATE_EOI_PASS
+        match eoi_decision:
+            case Decision.FAIL:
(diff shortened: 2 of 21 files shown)


.vscode/settings.json
```diff
@@ -16,5 +11,10 @@
       },
       "editor.formatOnSave": true,
       "editor.defaultFormatter": "charliermarsh.ruff"
-    }
+    },
+    "python.testing.pytestArgs": [
```
gidsg, line 15: These changes are unrelated so could be in a separate PR?

application_store/_helpers/application.py
```diff
@@ -19,3 +33,57 @@ def order_applications(applications, order_by, order_rev):
...
+                application.round_id,
+            ).get("lead_contact_name", "")
+        )
+
+        if Decision(eoi_decision) == Decision.PASS:  # EOI Full pass
+            notify_template = Config.NOTIFY_TEMPLATE_EOI_PASS
+
+        elif Decision(eoi_decision) == Decision.PASS_WITH_CAVEATS:  # EOI Pass with caveats
+            notify_template = Config.NOTIFY_TEMPLATE_EOI_PASS_W_CAVEATS
+            contents = [NotifyConstants.APPLICATION_CAVEATS] = (eoi_results["caveats"],)
```
gidsg, line 71: Could we refactor this using [Structural Pattern Matching](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching)?

application_store/_helpers/application.py
```diff
@@ -19,3 +33,57 @@ def order_applications(applications, order_by, order_rev):
...
+    else:
+        notify_template = Config.NOTIFY_TEMPLATE_SUBMIT_APPLICATION
+        eoi_decision = None
+        full_name = account.full_name
+
+    contents["application"] = create_qa_base64file(contents.get("application"), True)
+    del contents["application"]["forms"]
```
gidsg, line 78: Rather than mutating here could we return the thing we want directly e.g.
`application_contents = create_qa_base64file(contents.get("application"), True)` and pass that to the `Notification`?

assessment_store/db/queries/assessment_records/_helpers.py
```diff
@@ -10,147 +10,189 @@
...
+        "funding_amount_requested": 0,
+        "asset_type": "No asset type specified.",
+        "language": application_json["language"],
+        "location_json_blob": {
+            "error": False,
+            "postcode": FIELD_DEFAULT_VALUE,
+            "county": FIELD_DEFAULT_VALUE,
+            "region": FIELD_DEFAULT_VALUE,
+            "country": FIELD_DEFAULT_VALUE,
+            "constituency": FIELD_DEFAULT_VALUE,
```
gidsg, line 118: Would prefer for this to be `null` and the frontend handle what to display. But OK if this is a lift and shift of the previous behaviour.

srh-sloan: Yeah it's just what was there before, didn't want to have to change the frontend as well

.vscode/settings.json
srh-sloan: This can just go actually - I had to remove and re-add and VS code has just put in a different place in the file.

uv.lock
```diff
@@ -325,6 +325,30 @@ uvicorn = [
     { name = "uvicorn", extra = ["standard"] },
 ]
 
+[[package]]
```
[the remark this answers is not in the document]
srh-sloan: I added `pytest-cov` when writing the tests so it's brought these in but happy to move to a different PR 👍

application_store/db/models/application/applications.py
```diff
@@ -42,7 +42,11 @@ class Applications(BaseModel):
     __table_args__ = (db.UniqueConstraint("fund_id", "round_id", "key", name="_reference"),)
 
     def as_dict(self):
-        date_submitted = self.date_submitted.isoformat() if self.date_submitted else "null"
+        date_submitted = (
+            (self.date_submitted if type(self.date_submitted) is str else self.date_submitted.isoformat())
```
samuelhwilliams, line 46: this feels odd - I think we should have this always as a datetime instance - do you know where the string version is coming from?

application_store/_helpers/application.py
srh-sloan: Have tweaked this slightly but reluctant to change any further as it's feeling less like a shift and shift and that function is used in other places.

application_store/_helpers/application.py
srh-sloan: Yep done

application_store/db/models/application/applications.py
samuelhwilliams: sorry yep ignore me - we still are reading from the queue for one deployment 👍

samuelhwilliams (dismissed):
have tested this locally and it works, so don't want to block too much ✅ 

But there is lots of juicy stuff here to consolidate in the next few weeks/months...

application_store/_helpers/application.py
```diff
@@ -19,3 +33,53 @@ def order_applications(applications, order_by, order_rev):
...
+        match eoi_decision:
+            case Decision.FAIL:
+                return  # we don't send emails for a failure
+
+            case Decision.PASS:
+                notify_template = Config.NOTIFY_TEMPLATE_EOI_PASS
+
+            case Decision.PASS_WITH_CAVEATS:
+                notify_template = Config.NOTIFY_TEMPLATE_EOI_PASS_W_CAVEATS
+                contents[NotifyConstants.APPLICATION_CAVEATS] = eoi_results["caveats"]
```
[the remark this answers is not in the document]
srh-sloan: Done

application_store/_helpers/application.py
```diff
@@ -19,3 +33,53 @@ def order_applications(applications, order_by, order_rev):
...
+    account,
+    application_with_form_json_and_fund_name,
+    application,
+    round_data,
+):
+    contents = {
+        NotifyConstants.APPLICATION_FIELD: create_qa_base64file(application_with_form_json_and_fund_name, True),
+        NotifyConstants.MAGIC_LINK_CONTACT_HELP_EMAIL_FIELD: round_data.contact_email,
+    }
+    del contents[NotifyConstants.APPLICATION_FIELD]["forms"]
```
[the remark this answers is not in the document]
srh-sloan: Added a unit test. Agree with the worry around `del` - it was already in there and is another thing we can properly refactor after the lift/shift queue removal

samuelhwilliams (approved):

PR #53, opened 2024-12-16
Remove nested account_store pre-commit config

samuelhwilliams:
And fix pre-commit issues that are currently on main


sfount (approved):
I'm not confident ruff is running consistently in the pre-commit checks even outside of this nest config but this definitely needs to happen either way!

srh-sloan (approved):

PR #58, opened 2024-12-17
FSPT-190 Shift account store 

sfount:
Removes anything from the nested `account_store` folder that is now managed by the top level pre-award-stores app. Adds the account store API spec as a new route and migrations based on the account store models.

Things to be aware of: 
- this PR doesn't set `filterwarnings` to error in the pytest.ini file, there are many warnings from our dependencies but I believe there is also one or two warnings from our direct code
- the `validate_responses=True` for adding a connexion API causes the endpoints for account store to fail, this PR just sets this to false but I suspect that comes to down the properties being incorrectly flagged as non-optional


app.py
```diff
@@ -51,6 +51,13 @@ def create_app() -> FlaskApp:
         resolver=MethodViewResolver("api"),
     )
 
+    connexion_app.add_api(
+        get_bundled_specs("/account_store/openapi/api.yml"),
+        validate_responses=False,
```
sfount, line 56: Note this defaults to `None` which will set it to false, this is in the spirit of lifting and shifting.

samuelhwilliams: i can't believe you've done this.gif

samuelhwilliams (approved):
yo....lo

(LGTM)

PR #59, opened 2024-12-17
adding script for errors in deriving values

srh-sloan:
### Change description
Now that we insert into assessment_store when a user submits an application, the `derive_values` call can fail. The user is not blocked but we write any such errors to Sentry.

This PR adds a script to manually run `derive_values` for a specific application, and then optionally commit those changes to the DB.

Run book for when and how to use this script: https://mhclgdigital.atlassian.net/wiki/spaces/FS/pages/81068033/Pre-Award+Common+Support+Tasks#Re-run-derive_values-for-an-application-in-assessment_store

### How to test
- Idenitfy an application that exists in `assessment_records`
- Update that application to have some arbitrary value in any of `funding_amount_requested`, `asset_type` or `location_json_blob`
- Run the script with `python -m assessment_store.scripts.derive_assessment_values -a <application_id>`
- After confirmation, you should see that row updated with correct values for the fields above

### DIFF
--- a/assessment_store/scripts/derive_assessment_values.py
+++ b/assessment_store/scripts/derive_assessment_values.py
@@ -1,0 +1,53 @@
+import click
+
+from application_store.db.queries.application.queries import get_application
+from assessment_store.db.models.assessment_record.assessment_records import AssessmentRecord
+from assessment_store.db.queries.assessment_records._helpers import derive_application_values
+
+
+@click.command()
+@click.option(
+    "-a",
+    "--application_id",
+    help="Application ID to derive values",
+    prompt=True,
+)
+# @click.option(
+#     "-c",
+#     "--commit",
+#     is_flag=True,
+#     default=False,
+#     show_default=True,
+#     help="Whether to commit the derived values to assessment_store",
+#     prompt=False,
+# )
+def derive_assessment_values(application_id):
+    from flask import current_app
+
+    from db import db
+
+    current_app.logger.info(
+        "Deriving values for {application_id}",
+        extra=dict(application_id=application_id),
+    )
+
+    application_json = get_application(application_id, include_forms=True, as_json=True)
+    derived_values = derive_application_values(application_json)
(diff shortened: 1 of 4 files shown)


assessment_store/scripts/derive_assessment_values.py
```diff
@@ -0,0 +1,53 @@
...
+    from db import db
+
+    current_app.logger.info(
+        "Deriving values for {application_id}",
+        extra=dict(application_id=application_id),
+    )
+
+    application_json = get_application(application_id, include_forms=True, as_json=True)
+    derived_values = derive_application_values(application_json)
+    current_app.logger.info("Derived the following values {derived_values}", extra=dict(derived_values=derived_values))
```
samuelhwilliams, line 36: could we pretty-print this rather than console log it maybe? Might make it easier to review the derived values if they're multi-line/indented, rather than a single long line?

assessment_store/scripts/derive_assessment_values.py
```diff
@@ -0,0 +1,53 @@
...
+        current_app.logger.info("Database updated with derived values")
+    else:
+        current_app.logger.info("Database not updated")
+
+
+if __name__ == "__main__":
+    from app import app
+
+    with app.app.app_context():
+        derive_assessment_values()
```
samuelhwilliams, line 53: Random-ish question - should we make this an invoke command rather than a script? Do we have a rationalise for using one over the other?

assessment_store/scripts/derive_assessment_values.py
srh-sloan: Agree, done

assessment_store/scripts/derive_assessment_values.py
srh-sloan: I guess because I knew how to use click to do the confirmation stuff? No real rationale, yes we should standardise

samuelhwilliams (approved):

PR #61, opened 2024-12-18
removing unused queue code

srh-sloan:
### Change description
Removing the code that reads from the queue in assessment, and a bunch of other scripts we don't need any more now we write directly to assessment store on submit.

- Removed copilot config for the assess import queue
- Companion PR for docker runner to remove that config from local https://github.com/communitiesuk/funding-service-design-docker-runner/pull/108
- Moved `bulk_insert_application_record` into `conftest` as it is only used by tests now (previously used by the queue reader)
- Removed unused config values from `DefaultConfig` and envs
- Removed assessment polling code, and associated methods
- Removed old `location_utils` and `import_applications` scripts
- Removed appropriate unit tests

### How to test
Pull this branch and the linked docker-runner branch
Submit an application. It should correctly appear in assessment, and the submit notifcation should be sent

None


tests/assessment_store_tests/conftest.py
```diff
@@ -38,6 +46,73 @@ def ctdf_application():
     yield copy.deepcopy(ctdf_input_data)
 
 
+def bulk_insert_application_record(
```
sfount, line 49: The other methods in here seem to work as `@pytest.fixtures` where they accept the database connection used by the running tests, do some things and yield a result. I think ideally this would work in the same way but if we think this will only be around for a short while this should be fine?

sfount (approved):
Looks good! Small question about carrying over the now test specific functionality but I don't think it should be blocking

I've not thoroughly checked that this removes every hint of queueing but it looks comprehensive! If there's anything remaining I'm confident it would come out of upcoming refactoring

tests/assessment_store_tests/conftest.py
srh-sloan: Yeah I didn't want to mess with that too much as I think we have a TODO around sorting out unit test data, and also this inserts into assessment which will hopefully soon be one and the same as apply

samuelhwilliams:
Do we also need a companion PR on workflows to remove the [assessment queue](https://github.com/communitiesuk/funding-service-design-workflows/blob/main/apps/pre-award/copilot/environments/addons/assessment-import-queue.yml)?

samuelhwilliams (approved):

PR #67, opened 2024-12-23
Fspt/remove connexion

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-201

### Change description
Removes `connexion` which hooks up `openapi.yml` files to Flask routes automatically. We're removing this so that we can, in short order, pull in `pre-award-frontend` and remove most of the internal API routes all together.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,4 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 ENV PATH="/app/.venv/bin:$PATH"
 EXPOSE 8080
 
-CMD ["gunicorn", "--worker-class", "uvicorn.workers.UvicornWorker", "wsgi:app", "-b", "0.0.0.0:8080"]
+CMD ["gunicorn", "--worker-class", "gevent", "wsgi:app", "-b", "0.0.0.0:8080"]
--- a/assessment_store/api/routes/flag_routes.py
+++ b/assessment_store/api/routes/flag_routes.py
@@ -1,4 +1,4 @@
-from flask import current_app, request
+from flask import Blueprint, current_app, request
 
 from assessment_store.db.models.flags.flag_update import FlagStatus
 from assessment_store.db.queries import get_metadata_for_fund_round_id
@@ -10,6 +10,8 @@ from assessment_store.db.queries.flags.queries import (
 )
 from assessment_store.db.schemas.schemas import AssessmentFlagSchema
 
+assessment_flag_bp = Blueprint("assessment_flag", __name__)
+
 
 def _fix_country(country):
     country = country.strip().casefold()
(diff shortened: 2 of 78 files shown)


tests/application_store_tests/test_routes.py
```diff
@@ -593,7 +596,7 @@ def test_form_data_save_with_closed_round(flask_test_client, seed_application_re
     response = flask_test_client.put(
         "/application/applications/forms",
         json=section_put,
-        follow_redirects=True,
+        follow_redirects=False,
```
sfount, line 599: 🤔 (going from following redirects to not feels like a safe way around)

assessment_store/api/routes/flag_routes.py
```diff
@@ -69,20 +73,24 @@ def create_team_dict(team_name):
     return team_flag_stats
 
 
-def get_flag(flag_id: str):
+@assessment_flag_bp.get("/flag_data")
+def get_flag():
+    flag_id = request.args["flag_id"]
```
sfount, line 78: As a reminder to myself, previous checking done by connexion (assuming it makes sure there's a flag id query string here) now _won't be done_ -- this should be fine if we're moving away from the stores exposing endpoints

assessment_store/openapi/api.yml
```diff
@@ -428,26 +428,6 @@ paths:
         schema:
           type: boolean
         required: false
-      - name: score
```
sfount, line 431: This seemed unused - the open api specs will drift from reality now as they're not generated based on the flask app. We should remove them when we come to cleanup

sfount (approved):
Really good work, plenty to remove as we go through but this will help us do it in smaller increments

Dockerfile
```diff
@@ -20,4 +20,4 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 ENV PATH="/app/.venv/bin:$PATH"
 EXPOSE 8080
 
-CMD ["gunicorn", "--worker-class", "uvicorn.workers.UvicornWorker", "wsgi:app", "-b", "0.0.0.0:8080"]
+CMD ["gunicorn", "--worker-class", "eventlet", "wsgi:app", "-b", "0.0.0.0:8080"]
```
sfount, line 23: As discussed https://eventlet.readthedocs.io/en/latest/#warning

sfount (changes requested):

Dockerfile
samuelhwilliams: Thanks for catching this Steven, really good find 🙏

We need to switch to _something_ other than Uvicorn, because that is designed for ASGI apps - which Connexion was. Native Flask is a WSGI app (and `gunicorn` is a WSGI server), so we have a few options for workers:

1) `sync` - fully synchronous, 1 request per worker at a time
3) `eventlet` - have generally seen this used across other government apps (eg Notify), but as you say the docs suggest not using it for new things. Had a bunch of issues getting monkeypatching happening early enough, and have been scarred a few times using eventlet in the past so I think we're good to avoid this.
2) `gevent` - also green thread based, should be a better option than `sync` and does seem to just be working without having to fiddle too hard.

So will go for `gevent`.

samuelhwilliams:
@sfount works and passes e2e tests in `tests` env - ready for re-review

sfount (approved):

PR #70, opened 2024-12-24
Serialize JSON in line with connexion

samuelhwilliams:
### Change description
Connexion serializes dates/datetimes differently from Flask, and pre-award-frontend can only deal with the Connexion `.isoformat()` serialization currently. This throws [an error](https://funding-service-design-team-dl.sentry.io/issues/6173529391).

We'll be removing this serialization soon, so the quickest fix is to keep that serialization format.

Source code:
https://github.com/spec-first/connexion/blob/main/connexion/jsonifier.py#L13-L46

This serialization change broke the feedback surveys, which pass back datetimes that are parsed by pre-award-frontend expecting a 'Z' suffix, eg
https://github.com/communitiesuk/funding-service-pre-award-frontend/blob/3ffbd7dfa63c4eb71553317334c9e2bd39dd59a7/apply/helpers.py#L354

### How to test
1. On current main, complete the crash test dummy fund application.
2. If you try and complete a feedback/research survey, it will crash.
3. Checkout this branch
4. Do the survey - should work fine.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,7 +1,9 @@
+import datetime
 from os import getenv
 
 import psycopg2
 from flask import Flask, jsonify
+from flask.json.provider import DefaultJSONProvider
 from fsd_utils import init_sentry
 from fsd_utils.healthchecks.checkers import DbChecker, FlaskRunningChecker
 from fsd_utils.healthchecks.healthcheck import Healthcheck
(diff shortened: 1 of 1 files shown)


sfount (approved):

gidsg (approved):
📅 👍 
Might be nice to add a test, but OK to have a TODO if priority is to get this fix out.

samuelhwilliams:
agree a test would be nice - will follow up.

PR #72, opened 2024-12-24
FSPT-179 Add host matching

sfount:
Adds host matching to the existing store APIs allowing different routes/
blueprints to serve different host headers. This sets the app up to be
able to continue to serve the store APIs as frontends are added on top.

The existing store tests should set their headers as if they're
accessing the API routes.

Note this just configures each stores separately as thats how they are
setup from combining the repositories. This will likely be streamlined
to one test client for tests as we refactor away the store + frontend
boundaries.

In order for this to work locally this depends on:
- [x] https://github.com/communitiesuk/funding-service-design-docker-runner/pull/112

### DIFF
--- a/copilot/fsd-pre-award-stores/manifest.yml
+++ b/copilot/fsd-pre-award-stores/manifest.yml
@@ -54,6 +54,7 @@ variables:
     from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-NotificationDeadLetterQueueURL
   ASSESSMENT_FRONTEND_HOST: "https://assessment.${COPILOT_ENVIRONMENT_NAME}.access-funding.test.levellingup.gov.uk"
   SENTRY_TRACES_SAMPLE_RATE: 1.0
+  API_HOST: "fsd-pre-award-stores:8080"
 
 secrets:
   FUND_STORE_API_HOST: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/FUND_STORE_API_HOST
--- a/tests/application_store_tests/conftest.py
+++ b/tests/application_store_tests/conftest.py
@@ -1,13 +1,17 @@
 from datetime import datetime, timedelta
+from typing import Any
 from uuid import uuid4
 
 import pytest
+from flask.testing import FlaskClient
+from werkzeug.test import TestResponse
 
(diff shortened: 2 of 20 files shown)


tests/application_store_tests/conftest.py
```diff
@@ -27,8 +31,25 @@ def app():
...
+            kwargs["headers"].setdefault("Host", Config.API_HOST)
+        else:
+            kwargs.setdefault("headers", {"Host": Config.API_HOST})
+        return super().open(*args, buffered=buffered, follow_redirects=follow_redirects, **kwargs)
+
+
 @pytest.fixture(scope="function")
-def flask_test_client(app):
+def flask_test_client():
+    app = create_app()
```
sfount, line 51: Moving this from being injected to re-instantiated every time didn't feel good but something about the test layout + order causes half of them to fail as it was configured.

copilot/fsd-pre-award-stores/manifest.yml
```diff
@@ -54,6 +54,7 @@ variables:
     from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-NotificationDeadLetterQueueURL
   ASSESSMENT_FRONTEND_HOST: "https://assessment.${COPILOT_ENVIRONMENT_NAME}.access-funding.test.levellingup.gov.uk"
   SENTRY_TRACES_SAMPLE_RATE: 1.0
+  API_HOST: "http://fsd-pre-award-stores:8080"
```
samuelhwilliams, line 57: No `http://` here please, I believe 🙏

samuelhwilliams: It's deceptive because I think you've looked at eg `FUND_STORE_API_HOST`, which will have `http://` - however those are used for building URLs, while this needs to be a `host` in URI terms.

FUND_STORE_API_HOST should probably be called something like FUND_STORE_API_BASE_URL or something equally zany-but-slightly-more-specific

tests/fund_store_tests/conftest.py
```diff
@@ -184,9 +188,26 @@ def app() -> Flask:
     yield app
 
 
+class _ApplyFlaskClient(FlaskClient):
```
samuelhwilliams, line 191: As discussed 👀

samuelhwilliams (approved):
Looks good 👍

PR #73, opened 2024-12-24
Make `PUT application tags` return a value if no tags

samuelhwilliams:
Because otherwise Flask throws an error


sfount (approved):
Because [connexion was doing this](https://github.com/communitiesuk/funding-service-pre-award-stores/pull/67) before, presumably

PR #74, opened 2024-12-24
Pull in pre-award-frontend

samuelhwilliams:
### Change description
Takes a cut of `pre-award-frontend` and pulls it into the `frontend` subdirectory so that we can start working on integrating it.

```
    Add 'frontend/' from commit '3ffbd7dfa63c4eb71553317334c9e2bd39dd59a7'

    git-subtree-dir: frontend
    git-subtree-mainline: 2e2397030c59d00981523bd02a0246a8a58bae37
    git-subtree-split: 3ffbd7dfa63c4eb71553317334c9e2bd39dd59a7
```

If any new commits go into `pre-award-frontend` then we'll need to update this and rebase any pending changes.

The integration work should happen as a single PR and `pre-award-frontend` should be archived just before merging the integration PR.

https://mhclgdigital.atlassian.net/browse/FSPT-204


sfount (approved):
every line +84,816

PR #78, opened 2024-12-27
Created DB changes and updated submit_application to allow resubmissi…

ksp37-dluhc:
### Change description
Backend and DB changes to enable resubmission of applications. This PR covers the following changes from the associated ticket:

- Extend Flags to include a list of affected fields and a boolean to indicate if it's a change request
- Add two new application statuses ‘Change received’ and ‘Change requested’  
- On resubmission of an application, the processing will
    - mark updated sub-criteria in json
    - store a diff of answers before and after changes
    - update derived values and application status 
    - resolve change request flags that have been addressed (should be all of them)

### How to test
N/A

N/A


db/migrations/versions/009_add_binary_scoring_system.py
```diff
@@ -0,0 +1,31 @@
...
+zero_to_one_id = "40d10744-b0c1-4af0-9cb9-df1058a3e930"
+
+
+def upgrade():
+    op.execute(
+        sa.text(
+            "INSERT INTO scoring_system (id, scoring_system_name, minimum_score, maximum_score) "
+            "VALUES (:id, 'ZeroToOne', 0, 1)"
+        ).bindparams(id=zero_to_one_id)
+    )
```
samuelhwilliams, line 27: Please can we not use alembic migrations for inserting data - this can cause problems over time and it often leads to code that is tightly intermingled with the expectation of certain data. I think we'd like to avoid this going forwards.

[Alembic migrations should be used for structural changes only, ideally](https://github.com/sqlalchemy/alembic/discussions/972#discussioncomment-1766394). Occasionally I think it _can_ be OK to do UPDATEs that affect existing data to wrangle them into the right shape, but nothing that adds new data from scratch.

A good rule of thumb to keep is that after all the alembic migrations have run, we should always have an empty database.

If we need to insert a new entry here, can we add a script or invoke task that gets manually run to insert the data for now please? 🙏

db/migrations/versions/009_add_binary_scoring_system.py
```diff
@@ -0,0 +1,31 @@
...
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "009_add_binary_scoring_system"
+down_revision = "008_extend_flag_for_change_req"
+branch_labels = None
+depends_on = None
+
+zero_to_one_id = "40d10744-b0c1-4af0-9cb9-df1058a3e930"
```
samuelhwilliams, line 18: I also think that we shouldn't be hardcoding IDs here, as that can lead to then just hardcoding the IDs in the codebase.

I think it's safer and leads to more robust code if the IDs are random in all environments. That means we then need to be able to pull the record out of the DB by some other identified, such as a descriptive name, and leads to better relational data and access patterns.

db/migrations/versions/009_add_binary_scoring_system.py
samuelhwilliams: Also - this scoring system should be added to the local seeding script here: https://github.com/communitiesuk/funding-service-pre-award-stores/blob/27e15e24de3d0e9fa5a2089ac19215a1d9c5c6fc/assessment_store/tasks/db_tasks.py#L127

That's how we get data into the local DB.

It doesn't solve the "how do we get data into the live DBs" - but we need to do that separately. I would suggest that until we build a robust answer, maybe just write a script that can be run against an environment to insert it?

ksp37-dluhc: Thanks for having a look at this! I was a bit unsure how we're approaching things in the "new monolithic world" so I just followed the existing patterns that we've been using.  

I do agree in principal that migrations aren't suited to adding data, but as in the discussion you linked, I'm not sure there is a definitively better alternative. Personally I think adding a scoring system is fairly safe since it's essentially seed data and really it shouldn't even have a downgrade (even though I've provided it, perhaps I shouldn't have). 

But I'm happy to create a script instead until we've had a chance to agree a long-term solution. Do we currently auto run any data scripts as part of deployment or is it something I'll need to do manually?

db/migrations/versions/009_add_binary_scoring_system.py
ksp37-dluhc: Yeah that's true! I've leave it to be autogenerated.

db/migrations/versions/009_add_binary_scoring_system.py
samuelhwilliams: We don't have anything that would auto run the data migrations, yeah - so it would just be a manual task. If you add a script somewhere then it should be a case of something like `copilot svc exec --name fsd-pre-award --command "launcher python <scriptname>` - so not super overly arduous (but you have to do it for each environment, and consider how you're deploying things so that the first deploy gets the script in the environment but can survive without the data, etc).

application_store/db/queries/application/queries.py
```diff
@@ -266,25 +269,96 @@ def submit_application(application_id) -> Applications:
...
+            stmt = postgres_insert(AssessmentRecord).values(row)
+
+            # setting with an update makes sure the derived values are recalculated
+            update_row_statement = stmt.on_conflict_do_update(
+                index_elements=[AssessmentRecord.application_id], set_=row
+            ).returning(AssessmentRecord.application_id)
+
+            db.session.execute(update_row_statement)
+
+            flags = db.session.scalars(
```
albertkol, line 317: Similar here. I think this relation is already defined, so it can be assessment_flag.flags.

If you want to apply filter on it, there are a number of ways to go about it:
1. You can fetch all and iff inside the for loop (bit nasty)
2. You can add a `hybrid_property` inside the model:
```
    @hybrid_property
    def unfinished_flags(self):
        return [flag for flag in self.flags if not flag.is_finished]
```
Still similar to 1. but the query is done inside the model.
3. (the most efficient one) define the flags relation as:
```
    flags = relationship(
        "AssessmentFlag",
        lazy="dynamic",  # Makes it queryable
    )
```
So you should be able to do something like this:
```
unfinished_flags = assessment_record.flags.filter_by(is_finished=False).all()
```

application_store/db/queries/application/queries.py
```diff
@@ -266,25 +269,96 @@ def submit_application(application_id) -> Applications:
...
         }
-        stmt = postgres_insert(AssessmentRecord).values([row])
 
-        upsert_rows_stmt = stmt.on_conflict_do_nothing(index_elements=[AssessmentRecord.application_id]).returning(
-            AssessmentRecord.application_id
+        existing_application = db.session.scalar(
+            select(AssessmentRecord)
+            .where(
+                AssessmentRecord.application_id == row["application_id"],
+                AssessmentRecord.is_withdrawn == False,  # noqa: E712
```
albertkol, line 277: Not too sure, but can we not avoid the typing problem by using the `not_` from `sqlalchemy`?

application_store/db/queries/application/queries.py
```diff
@@ -266,25 +269,96 @@ def submit_application(application_id) -> Applications:
...
+                                field["history_log"].append(
+                                    {datetime.now(tz=timezone.utc).isoformat(): field_map[field["key"]]}
+                                )
+                            else:
+                                field["history_log"] = [
+                                    {datetime.now(tz=timezone.utc).isoformat(): field_map[field["key"]]}
+                                ]
+
+            if changed_fields:
+                row["workflow_status"] = WorkflowStatus.CHANGE_RECEIVED
```
albertkol, line 306: This one is up to you... I don't like `helpers` functions very much, but this update row logic might be nice if extracted so it can be tested individually.

application_store/db/queries/application/queries.py
```diff
@@ -266,25 +269,96 @@ def submit_application(application_id) -> Applications:
...
+
+            upsert_rows_stmt = stmt.on_conflict_do_nothing(index_elements=[AssessmentRecord.application_id]).returning(
+                AssessmentRecord.application_id
             )
-        db.session.commit()
+
+            result = db.session.execute(upsert_rows_stmt)
+
+            # Check if the inserted application is in result
+            inserted_application_ids = [item.application_id for item in result]
```
albertkol, line 352: Not sure what this achieves. Isn't the if below only len(result)?

assessment_store/db/models/flags/flag_update.py
```diff
@@ -26,7 +26,7 @@ class FlagUpdate(BaseModel):
         UUID(as_uuid=True),
         ForeignKey("assessment_flag.id"),
     )
-    user_id = Column("user_id", String, nullable=False)
+    user_id = Column("user_id", String, nullable=True)
```
albertkol, line 29: Can this be nullable?

assessment_store/db/queries/flags/queries.py
```diff
@@ -39,6 +41,8 @@ def add_flag_for_application(
         latest_allocation=allocation,
         latest_status=status,
         updates=[flag_update],
+        field_ids=field_ids if field_ids else [],
```
albertkol, line 44: Up to you, I'd leave it `None` if does not exist.

application_store/db/queries/application/queries.py
samuelhwilliams: or maybe `is False`, if that works (I'm not sure)?

application_store/db/queries/application/queries.py
```diff
@@ -240,14 +243,14 @@ def search_applications(**params):
...
-def submit_application(application_id) -> Applications:
+def submit_application(application_id) -> Applications:  # noqa: C901
     current_app.logger.info(
         "Submitting application {application_id} and importing to assessment store",
         extra=dict(application_id=application_id),
     )
     try:
         application = get_application(application_id, include_forms=True)
-
+        fund = get_fund(application.fund_id)
```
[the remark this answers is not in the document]
ksp37-dluhc: I wish we could, but `application.fund_id` isn't actually a foreign key to `Fund` (one used to be in application-store and the other in fund-store, so they don't have referential integrity).

assessment_store/db/models/flags/flag_update.py
ksp37-dluhc: Originally a user had to resolve a flag, but now in `submit_application`, we automatically resolve flags that are change requests and have been addressed. So there's no "user" to specify in that situation.

samuelhwilliams: Is the user in this case maybe the person submitting the application?

application_store/db/queries/application/queries.py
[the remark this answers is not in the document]
samuelhwilliams: Are we able to add referential integrity now?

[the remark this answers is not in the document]
ksp37-dluhc: I think so, but there's a few places where it should be done so probably better in a dedicated PR rather than bit part.

assessment_store/db/models/flags/flag_update.py
ksp37-dluhc: Good point! Completely missed that

db/migrations/versions/009_add_binary_scoring_system.py
ksp37-dluhc: @hamzabinkhalid is adding the scoring system as part of his PR, so I've removed it from mine and referred him to this discussion.

application_store/db/queries/application/queries.py
ksp37-dluhc: I ended up adding a new relationship `change_requests` that filters on the join - I think that'll come in handy for the work we're doing.

application_store/db/queries/application/queries.py
ksp37-dluhc: I think it didn't recognise that it's legitimate SQLAlchemy syntax (https://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause) 

...anyway I bended to the will of the linter and used a workaround.

application_store/db/queries/application/queries.py
ksp37-dluhc: Yeah good shout - I've moved it out into a separate function and added some tests.

application_store/db/queries/application/queries.py
ksp37-dluhc: Yep - this was from the existing code but I didn't notice it; updated it now.

albertkol: 🤯

application_store/db/queries/application/queries.py
```diff
@@ -240,14 +242,37 @@ def search_applications(**params):
...
+                        field["history_log"].append(
+                            {datetime.now(tz=timezone.utc).isoformat(): field_map[field["key"]]}
+                        )
+                    else:
+                        field["history_log"] = [{datetime.now(tz=timezone.utc).isoformat(): field_map[field["key"]]}]
+
+    return changed_fields
+
+
+def submit_application(application_id) -> Applications:  # noqa: C901
```
albertkol, line 268: Not the end of the world. But with the recent `update_application_fields` extraction, do we still break the C901  rule?

albertkol (approved):
I think we can ship it.. 🧊 🚢 💥 

The resolution of flags after re-submitting sections that received feedback are yet to be confirmed if the work.

PR #80, opened 2024-12-27
FSPT-179 Rename README

sfount:
This will need to be updated again when we've completed refactoring as we're moving away from terminology like pre-award but this removes the store specific language to line it up with local setups and deployments.


albertkol (approved):

samuelhwilliams (approved):
:chefskiss: ✅

PR #84, opened 2024-12-28
Re-enable DB migrations checking

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-204

### Change description
We temporarily disabled DB migrations checking for the first deployment of the new combined `funding-service-pre-award` app. Now that deployment has hit prod, we need to turn this back on.


albertkol (approved):
:+1

PR #88, opened 2024-12-30
Fix running individual tests

samuelhwilliams:
The `mock_redis` fixture was being setup incorrectly - before app creation. One of the things that mock_redis mocks is a method on `app.redis_mlinks`, which is only fully configured *after* the app has been made.


albertkol (approved):
Brill work!

PR #89, opened 2024-12-30
BAU Fix host matching for authenticator errors

sfount:
`AUTHENTICATOR_HOST` looks like it refers to the publicly addressable URL for authenticator (used to build links), `AUTH_HOST` is the internal server name part we'd expect to be passed in the request header.

This confusion should only exist in the short period the apps and stores are combined but haven't had the network layer removed/ been de-duplicated.


samuelhwilliams (approved):

samuelhwilliams:
good spot 👀

samuelhwilliams (approved):

PR #90, opened 2024-12-30
Add Flask-DebugToolbar

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-170

This provides some nice introspective utilities for the app when running on local developer machines, eg log messages emitted, templates rendered, and SQL queries executed.

SQL queries won't be visible yet because most API calls are still done over HTTP - but when we start to remove that abstraction, they'll show up.

## Show it
image

### DIFF
--- a/app.py
+++ b/app.py
@@ -8,6 +8,13 @@ from flask.json.provider import DefaultJSONProvider
 from flask_assets import Environment
 from flask_babel import Babel, gettext, pgettext
 from flask_compress import Compress
+
+try:
+    from flask_debugtoolbar import DebugToolbarExtension
+
+    toolbar = DebugToolbarExtension()
+except ImportError:
+    toolbar = None
 from flask_redis import FlaskRedis
 from flask_session import Session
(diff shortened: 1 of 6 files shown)


app.py
```diff
@@ -8,6 +8,13 @@
 from flask_assets import Environment
 from flask_babel import Babel, gettext, pgettext
 from flask_compress import Compress
+
+try:
+    from flask_debugtoolbar import DebugToolbarExtension
+
+    toolbar = DebugToolbarExtension()
```
sfount, line 15: I wonder if we should also only set it if the flask env isn't production - our dependency management is much more consistent now and I'm not sure what the failure mode is but that would also be consistent with a few other gated checks

sfount (approved):
Nice! might be worth sharing this & how it can be useful at TCOP, especially as it starts tracking everything

sfount (dismissed):

sfount (approved):

PR #91, opened 2024-12-30
Call GOV.UK Notify directly for magic link emails

samuelhwilliams:
<table><tr><td><details><summary>This PR is part of a stack created with <a href="https://github.com/aviator-co/av">Aviator</a>.</summary>

* **#101**
* **#100**
* **#95**
* ➡️ **#91**
* `main`
</details></td></tr></table>

https://mhclgdigital.atlassian.net/browse/FSPT-172

### Change description
Start replacing SQS messages to the notification service with simple synchronous API calls to the GOV.UK Notify API. The technical complexity and overhead added by having a distributed messaging queue and separate service is currently not proving to be a good value proposition.

This PR moves across the 'send magic link' email functionality - other emails remain going through the queue for now.

Writing new code from scratch feels easier, in this case, than trying to lift+shift+rationalise, because our current notifications service only handles 4 types of email.

The only part that *is* an intentional lift+shift is the fund configuration that is injected into the NotificationService. I anticipate rationalising this later (eg by using values from the DB instead).


config/envs/default.py
```diff
@@ -504,3 +504,5 @@ class DefaultConfig(object):
     # You can find the proper permission names from this document
     # https://docs.microsoft.com/en-us/graph/permissions-reference
     MS_GRAPH_PERMISSIONS_SCOPE = ["User.ReadBasic.All"]
+
+    GOV_NOTIFY_API_KEY = getenv("GOV_NOTIFY_API_KEY")
```
sfount, line 508: Do we want the app to start without this (in production and/ or locally)?

I think currently the magic link step is skipped if running locally but I'm not sure if it's skipped + doesn't add anything to the queue or if it sends off the email SQS message anyway (in which case it wouldn't get through without a valid key)

I haven't thought this through at all and happy to chat about it but I don't think having a valid Notify key should be a requirement for running locally (you just wouldn't get emails) -- I think that's an accidental property of having it distributed through a queue but it would be good to explicitly make the decision either way somewhere in these changes

services/notify/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+    def __init__(
+        self,
+        message="There was a problem sending the email through GOV.UK Notify",
+    ):
+        self.message = message
+        super().__init__(self.message)
+
+
+class NotificationService:
+    MAGIC_LINK_TEMPLATE_ID = os.environ.get("MAGIC_LINK_TEMPLATE_ID", "02a6d48a-f227-4b9a-9dd7-9e0cf203c8a2")
```
sfount, line 26: I'm not sure about having default values sprinkled through the code, does this want to live in the `envs/default` file?

To be honest I don't think we should have a default value for this at all in the code but appreciate that's what happens now - not a blocker

sfount: I've just looked further down and realised everything is doing this fetch from env + provide a default value, this can probably be tackled as a separate piece of work but none of it feels like the role of the notification service

sfount (approved):
Really thorough implementation of the flask service and nice assertions to be sure everything is correctly mapped.

I think most of my comments are about the logic for managing templates rather than the implementation here so I don't think any of them should block.

Great work

config/envs/default.py
samuelhwilliams: I do generally think we should hard-fail if our essential configuration is missing (see long-lived post-award ticket https://mhclgdigital.atlassian.net/browse/FPASF-163 👀) I think I'd rather do that change wholesale?

I probably agree that you shouldn't need a valid Notify key locally - which is what that config flag about not sending emails is for 🫣 There's also something around whether we should have a running mock notify service ala localstack. I wish Notify supplied something like that tbh, but maybe they're just find with people using test notify keys?

services/notify/__init__.py
samuelhwilliams: I'm on the fence - happy to discuss this but also have, as you realise, intentionally left this "for the future".

config/envs/default.py
sfount: Yeah great point there is an existing flag to do that. We can always revisit these threads when we're closer to the target code layout we want to arrive at.

Nice we could almost copy that post-award ticket wholesale

PR #97, opened 2024-12-31
Assess frontend

albertkol:
Adds the frontend on assess to:
- Accept questions (just the button) - this is to be used for the 1/0 scoring
- Request changes - GET and POST endpoints

Things left to be done:
- Set sub-criteria status to Changes Requested if they exist
- Display changes requested flag differently

### How to test
- pull branch locally
- `make pre up`
- Create an application on Apply f.x. CTDF application Pro tip: use lead_assessor example email
- Go to assess -> application -> subcriteria
- You will see the Accept/Request change button

### Screenshots
Accept / Request changes buttons:
image

Request changes form:
image

Display change requests on assess:
image

Display CHANGE_REQUEST status on tasklist:
image

Display CHANGE_RECEIVED status on tasklist:
image

Display CHANGE_REQUEST and CHANGE_RECEIVED statuses on applications page:
image


ksp37-dluhc:
Could you add a screenshot of the request changes page?

assess/assessments/routes.py
```diff
@@ -1172,6 +1175,19 @@ def display_sub_criteria(
     state = get_state_for_tasklist_banner(application_id)
     flags_list = get_flags(application_id)
```
ksp37-dluhc, line 1176: What do you think about having a method `get_change_requests()`? Internally it will call `get_flags()` for now but will do the filtering to only return change request flags. Similarly with `submit_change_request()` and `submit_flag()`. 

It means sometimes we'll have to make two API calls in place of one (at least until we move to DB queries directly), but the abstraction is useful to separate change requests from flags and save us having to do the filtering at the route level. It'll also save us some pain if we decide to move away from using flags for change requests in future.

assess/shared/helpers.py
```diff
@@ -82,6 +82,16 @@ def determine_flag_status(Flags: List[Flag]) -> str:
     flags_list = [(Flag.from_dict(flag) if isinstance(flag, dict) else flag) for flag in Flags] if Flags else []
     all_latest_status = [flag.latest_status for flag in flags_list]
 
+    has_flag_with_raised_status = False
```
ksp37-dluhc, line 85: I was thinking that when `request_changes()` is called, the workflow status should be updated to "Change requested" rather than having the logic as part of the flag statuses. Then `determine_display_status()` will pick up the status.

The behaviour is a bit different because the hierarchy goes flag status, then QA complete, then workflow status. So any flag status take precedence over the workflow status, but I think that's ok.

assess/assessments/routes.py
albertkol: Made the change.. and a bit more:
- `get_flags` is now only returning non change requests
- added `get_change_requests` function which only returns change requests
- created a simplified `submit_change_request` from the `submit_flag` as I did not need a lot of the mess that's in there anyway

I am not entirely sold on the idea that Flag (a.k.a. feedback for an assessor to an assessor) cannot co-exist with Change Requests (a.k.a. feedback from an assessor to an applicant), but this might be a safe approach for now.

assess/shared/helpers.py
albertkol: I agreed with your suggestion, and I think Flags should do the same: update the AssessmentRecort.workflow status every time, and have it be the source of truth without all these dynamic logics.

I made the change only for Change Requests to update the status. But it was not as clean as we hope it'd be. 

There is no endpoint to update an `AssessmentRecort`. So I ended up hitting the model directly. cc @samuelhwilliams I think creating a new endpoint or even modifying an existing one feels worse of an option.

albertkol:
Updated **How to test** section and added **screenshots**.

assess/templates/assess/macros/criteria_element.html
```diff
@@ -3,6 +3,19 @@
 
     {% set rows = [] %}
     {% for sub_criteria in criteria.sub_criterias %}
+        {% set status %}
```
ksp37-dluhc, line 6: Is anything being set here ..?

ksp37-dluhc (approved):

assess/shared/helpers.py
albertkol: We'll raise an RFC and refactor this later if the approach is unsatisfactory.

PR #99, opened 2024-12-31
fix failing tests

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-897

Fixed the tests that were failing locally by updating the function to correctly convert UTC to local time, ensuring the right timezone adjustments.

### DIFF
--- a/assess/shared/filters.py
+++ b/assess/shared/filters.py
@@ -1,7 +1,7 @@
 import ast
 from datetime import datetime
 
-from pytz import timezone
+from pytz import timezone, utc
 
 
 def slash_separated_day_month_year(value: str):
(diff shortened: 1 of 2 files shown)


assess/shared/filters.py
```diff
@@ -34,12 +34,15 @@ def utc_to_bst(value, tz="Europe/London", export_format=False):
     else:
         raise ValueError("Invalid datetime format")
 
-    uk_time = utc_time.astimezone(timezone(tz))
+    if utc_time.tzinfo is None:
+        utc_time = utc.localize(utc_time)
```
albertkol, line 38: Small thing, but instead of defaulting to `utc`, default to `tz`'s timezone. Which it's likely still UTC, but at least it takes into account if tz is provided.

```python
default_timezone = timezone(tz)
utc_time = default_timezone.localize(utc_time)
```

albertkol: Nvm.

albertkol (approved):
🚢 it

PR #104, opened 2025-01-02
Change API host for scheduled jobs

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-207

Change the host that API endpoints are exposed on from just the copilot app name (fsd-pre-award), to the app name plus service discovery namespace (fsd-pre-award.<env>.pre-award.local).

This is because we have a scheduled ECS task for post-award that talks to the account-store. Scheduled tasks are not able to resolve just the app's short name and need the namespace as well, so it would make requests to fsd-pre-award.dev.pre-award.local.

If we only configure the API endpoints' host_matching setting as `fsd-pre-award`, then those requests will 404. So we either need to support _both_ fsd-pre-award and the FQDN on API endpoints (more complex code change), or just switch everything to use the FQDN. This does the latter.

This deployment needs to be done in sync with updates to all of the FSD_PRE_AWARD_*_STORE_API_HOST parameter store variables.

### DIFF
--- a/account_store/core/account.py
+++ b/account_store/core/account.py
@@ -69,7 +69,10 @@ def get_bulk_accounts() -> Dict:
     :return:
         Nested dict of account_id: {account object}
     """
-    account_id = request.args.getlist("account_id")
+    raw_account_id = request.args.getlist("account_id")
+
+    # Filter out any empty/blank strings like '', which could happen with eg `?account_id=`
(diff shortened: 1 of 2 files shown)


albertkol (dismissed):
asd

albertkol (approved):

PR #105, opened 2025-01-02
FIX Parity with previous application date submitted format

sfount:
This fix formats the `date_submitted` property associated with assessment records using the format that has been used by all applications so far.

When the process moved from using a queue to directly updating the stores it somehow dropped the iso format meaning the timezone information is now seralised.

Reference https://github.com/communitiesuk/funding-service-pre-award/commit/46151b79e3652fcaefead063eeebac17df059a76

The assessment data export tool falls apart as it doesn't expect to see timezone information.


sfount:
Looking at this (quite dense) code again it looks like the iso format was moved to the application being serialised as a dictionary elsewhere, it should still be isoformat in theory -- will dig out where it might not be lining up

PR #107, opened 2025-01-03
Fix Applicant info export

sc15zs:
Description:
Fix Applicant info export error on Assess. 
Changed the date string parsing to dateutil.parser, which can parse strings both with or without timezone information.


assessment_store/db/queries/assessment_records/queries.py
```diff
@@ -741,8 +742,7 @@ def get_export_data(  # noqa: C901 - historical sadness
     final_list = []
 
     for assessment in assessment_metadatas:
-        iso_format = "%Y-%m-%dT%H:%M:%S.%f"
-        iso_datetime = datetime.strptime(assessment.jsonb_blob["date_submitted"], iso_format)
+        iso_datetime = parser.isoparse(assessment.jsonb_blob["date_submitted"])
```
sfount, line 745: Nice! Feels like a clean solution to get this back up and running

sfount (dismissed):
Looks good! Maybe we can add a version of `test_get_application_fields_export` in `assessment_store_tests/test_routes` that asserts that it will export with the two different timestamp formats?

sfount:
as we'll want this working in case there is an absence of an automated assertion I have run this locally and it resolves the described problem

sc15zs:
@sfount 
> Looks good! Maybe we can add a version of `test_get_application_fields_export` in `assessment_store_tests/test_routes` that asserts that it will export with the two different timestamp formats?

Added a unit test.

tests/assessment_store_tests/test_routes.py
```diff
@@ -446,6 +449,48 @@ def test_get_application_fields_export(flask_test_client, seed_application_recor
...
+
+    # Date strings with and without timezone information
+    date_strings = [
+        "2023-10-10T14:48:00+00:00",  # UTC with explicit timezone
+        "2023-10-10T14:48:00",  # UTC without explicit timezone
+        "2023-10-11T10:30:00+00:00",
+        "2023-10-11T10:30:00",
+    ]
+    for record, date_string in zip(assessment_records, date_strings, strict=False):
+        record.jsonb_blob["date_submitted"] = date_string
```
sfount, line 466: I assume these updates are processed and committed to the database synchronously? (I'm not too familiar with the sql alchemy session but just checking there isn't the chance that these end up in a transaction that waits for the end of the test to commit)

sfount: I think ideally the dates would be passed in through the `apps_to_insert` decorator so we wouldn't have to fetch them out of the database again before updating/ asserting but by asserting on the actual date values we can be confident that what has been updated is what was used so I think this should work!

sfount (approved):
Thanks for adding the coverage, confident in the change! 

We'll be doing a bit of thinking about how test data is seeded for integration tests so these will be included

tests/assessment_store_tests/test_routes.py
sc15zs: The date updates are not committed to the database. The test just fetches the 4 available test AssessmentRecord objects, and updates their _date_submitted_ property. 
Then it calls the _get_export_data_ function directly, in which the date strings are being parsed. 
Then it just compares the actual parsed values with the expected values. So, there's nothing committed to the database, it's just used within this function, and not saved anywhere.

PR #118, opened 2025-01-07
Fix renovate config

samuelhwilliams:
This should be pointing at github-hosted config, not whatever `local` is.

https://docs.renovatebot.com/config-presets/#github


MarcUsher (approved):
Oooh good spot

PR #121, opened 2025-01-07
Adding welsh fund config

srh-sloan:
Adding config for a new dummy fund 'Fund for Wales' that has welsh available.
Relies on [this PR](https://github.com/communitiesuk/digital-form-builder-adapter/pull/101) that adds the corresponding welsh form jsons


samuelhwilliams (approved):
Looks good locally 👍 

Should we add an e2e test for a Welsh-translated fund? 🤔

PR #123, opened 2025-01-07
Fix rendering of assessment PDF

samuelhwilliams:
There were a couple of issues with this that we fix here:

* The path to the template to render (full_application.html) has been broken since we consolidate repos. This fixes the path, and removes some of the custom jinja environment handling because that isn't needed any more.
* Update a couple of path references in docstrings and default CLI args.

## To test
1. Submit an application
2. Go to the assessment
3. Visit the `/export` URL, eg https://assessment.levellingup.gov.localhost:3010/assess/application/3dc45a6c-8284-4b94-b90d-81b593a8cd54/export
4. Click 'download PDF'

### DIFF
--- a/fund_store/scripts/generate_all_questions.py
+++ b/fund_store/scripts/generate_all_questions.py
@@ -152,13 +152,13 @@ def print_html(sections: dict) -> str:
 )
 @click.option(
     "--output_location",
-    default="../funding-service-design-frontend/app/templates/all_questions/",
+    default="../apply/templates/apply/all_questions/",
     help="Folder to write output html to (language code will be appended as an intermediate path)",
     prompt=True,
 )
 @click.option(
     "--forms_dir",
-    default="../digital-form-builder/fsd_config/form_jsons/",
+    default="../../digital-form-builder/fsd_config/form_jsons/",
     help="Local, absolute path, to the form JSONs to use to generate question lists",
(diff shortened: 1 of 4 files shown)


fund_store/scripts/generate_all_questions.py
```diff
@@ -152,13 +152,13 @@ def print_html(sections: dict) -> str:
...
     "--output_location",
-    default="../funding-service-design-frontend/app/templates/all_questions/",
+    default="../apply/templates/apply/all_questions/",
     help="Folder to write output html to (language code will be appended as an intermediate path)",
     prompt=True,
 )
 @click.option(
     "--forms_dir",
-    default="../digital-form-builder/fsd_config/form_jsons/",
+    default="../../digital-form-builder/fsd_config/form_jsons/",
```
sfount, line 161: is this referring to the relative code checked out? 😓 I think its `digital-form-builder-adapter` now

albertkol (approved):
:+1: I works indeed, although I did not read your To Test steps. I did not see them. Fix works!

sfount (approved):
I've not run this locally but confident the paths are now correct based on the apps package loaders. I'm not sure where/if that script is still run but theres quite a few things that are out of date. I wonder if we need a similar audit of scripts now that things are together

PR #124, opened 2025-01-07
Support python's datetime.isoformat in utc_to_bst

samuelhwilliams:
Some +00:00 datetime strings have slipped into our application/assessment 'date_submitted' fields, so we need to support those for now.

A more comprehensive fix should follow...


tests/assess_tests/test_filters.py
```diff
@@ -129,3 +130,36 @@ def test_remove_dashes_underscores_capitalize_keep_uppercase(self, address, expe
...
+        # Summer time / BST
+        ("2024-07-10T10:00:00.000000", "10/07/2024 at 11:00"),
+        ("2024-07-10 10:00:00.000000", "10/07/2024 at 11:00"),
+        ("2024-07-10 10:00:00", "10/07/2024 at 11:00"),
+        ("2024-07-10T10:00:00", "10/07/2024 at 11:00"),
+        ("2024-07-10 10:00:00.000000+0000", "10/07/2024 at 11:00"),
+        ("2024-07-10 10:00:00.000000+0100", "10/07/2024 at 10:00"),
+        ("2024-07-10 10:00:00.000000+00:00", "10/07/2024 at 11:00"),
+        ("2024-07-10 10:00:00.000000+01:00", "10/07/2024 at 10:00"),
+        ("2024-07-10T10:00:00.000000+00:00", "10/07/2024 at 11:00"),
```
sfount, line 159: 👍

sfount (approved):
Nice additional coverage and feels like a much more stable implementation - having less places doing serialisation/ deserialisation feels like it will be a big win for things like this

PR #125, opened 2025-01-07
FSPT-172 Send assessment emails

srh-sloan:
### Change description
Moves the 2 emails sent from assessment over to use the new synchronous notification pattern:
- assessment assigned
- assessment unassigned

### How to test
This functionality is currently behind a feature toggle that is disabled in production. I've enabled this locally and on dev to test, screenshots of sent emails below

Screenshot 2025-01-09 at 11 57 39
Screenshot 2025-01-09 at 11 57 51
Screenshot 2025-01-09 at 11 58 41


services/notify/__init__.py
```diff
@@ -347,6 +347,33 @@ def send_application_deadline_reminder_email(
             email_reply_to_id=self.REPLY_TO_EMAILS_WITH_NOTIFY_ID.get(contact_help_email),
         )
 
+    def send_assessment_email(
+        self,
+        template_id: str,
```
samuelhwilliams, line 352: One thing I tried to do with the notificationservice was have a single entrypoint for each semantic email, and that function would handle choosing the correct template in Notify (which might just be a single template or flip based on eg language).

I think it's nice to isolate all of the decisions around templates to this file, which we lose if we accept a `template_id`.

My justification for this is that the template you're using determines the personalisation that you need to send, and so deciding the template ID should be co-located with describing the personalisation - ie inside this function.

I think allowing any arbitrary template_id to be passed in breaks that contract slightly - it's not true to say that you can pass in any template ID and still send an assessment email.

So maybe we can move the template ID selection logic in here, and possibly drop all of `send_notification_email_assigned`, `send_notification_email_unassigned`,  and `_send_notification_email` from data_services.py by replacing calls to those to just direct calls to `get_notification_service().send_assessment_assigned_email` and `get_notification_service().send_assessment_unassigned_email`?

tests/assessment_store_tests/test_services.py
```diff
@@ -88,10 +83,23 @@ def test_send_notification_email_failure(
...
-    mock_logger.info.assert_called_with(
+    mock_logger.error.assert_called_with(
         "Could not send email for template: {template}, user: {user_id}, application {application_id}",
-        extra={"template": "assignment_template", "user_id": "user1", "application_id": "app1"},
+        extra={
+            "template": NotificationService.ASSESSMENT_APPLICATION_ASSIGNED,
+            "user_id": "user1",
+            "application_id": "app1",
+        },
+        exc_info=notification_error,
```
samuelhwilliams, line 104: Realise that this test was doing this kind of log capture before, but are we able to use the `caplog` fixture instead? I think it's a _generally_ more standard pytest way of doing this

services/notify/__init__.py
srh-sloan: Yes, I thought about this for ages because I figured that's the contract you'd intended with that `NotificationService`. But the 2 functions `send_assessment_assigned_email` and `send_assessment_assigned_email` both do quite a bit of retrieving fund and account info, and I didn't want to duplicate that

srh-sloan: Will take another look this morning and think some more

tests/assessment_store_tests/test_services.py
srh-sloan: Will have a look at this

samuelhwilliams (approved):
Functionally looks good to me, have confirmed the assign and unassign emails 👍 

image

image

nice one ✅

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #131, opened 2025-01-08
Fls 901 update eligibility navigation flow

hamzabinkhalid:
https://mhclgdigital.atlassian.net/browse/FLS-899
https://mhclgdigital.atlassian.net/browse/FLS-901

### Change description
- On MagicLink landing page show either "Continue" button to take the user to the fund round page to view existing applications if they have existing applications otherwise show a "Start a new application" button that redirects them to the eligibility form
- Redirect to tasklist after eligibility form is submitted

- [X] Unit tests and other appropriate tests added or updated
- [X] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
- Button Text Change:
        Login with Existing Applications: Use an email that already has existing applications. The MagicLink landing page should display a "Continue" button.
        Login with No Existing Applications(New user): Use an email with no existing applications. The MagicLink landing page should display a "Start a new application" button.
        Button Functionality:
            "Start a new application" button should take the user directly to the eligibility form.
            "Continue" button should take the user to the fund round dashboard, showing all existing applications and a button to start a new application.
        Eligibility Configuration: Ensure this change only applies to fund rounds with eligibility (eligibility_config: {has_eligibility: true}) in the round table, for which eligibility configs are enabled

- Direct Navigation to Tasklist Page:
        After completing the eligibility form, the user should be taken directly to the tasklist of the specified round.
        Ensure the previous behavior (showing a page indicating eligibility and requiring a button click to proceed to the tasklist) is replaced with direct navigation to the tasklist.

https://github.com/user-attachments/assets/c79a2dd1-7a51-4e42-b3fd-988cac993959


authenticator/frontend/templates/authenticator/partials/eligibility-check-button.html
```diff
@@ -0,0 +1,13 @@
+{%- from "govuk_frontend_jinja/components/button/macro.html" import govukButton -%}
+
+{% macro eligibilityCheckButton(form, url, fund_id, round_id) %}
```
sfount, line 3: It doesn't look like this is used?

authenticator/frontend/magic_links/routes.py
```diff
@@ -80,6 +82,21 @@ def landing(link_id):
...
+            if link_hash and isinstance(link_hash, bytes):
+                account_id = json.loads(link_hash.decode("utf-8"))["accountId"]
+                search_params = {
+                    "account_id": account_id,
+                }
+            elif g.is_authenticated:
+                search_params = {
+                    "account_id": g.account_id,
+                }
+            previous_applications = get_applications_for_account(**search_params)
```
sfount, line 98: This isn't directly related to this PR but it makes a bit nervous that it doesn't seem like theres anything on this end point that requires the code to pass any values (like account id, in this instance where there's a possibility its not set), if they're not set it seems like it would just return unfiltered applications which could easily trip us up. 

Just something we can keep in mind as we do any refactoring.

authenticator/api/magic_links/routes.py
```diff
@@ -64,6 +70,25 @@ def get(self, link_id: str):
 
             # Check link is not expired
             if link.get("exp") > int(datetime.now().timestamp()):
+                if round_data.has_eligibility:
+                    search_params = {
+                        "account_id": link.get("accountId"),
+                    }
+                    has_previous_applicaitons = get_applications_for_account(**search_params)
```
sfount, line 77: From a code structure point of view I wonder if further tying together the code that does authentication and business logic from the app might lead to some confusion further down the line?

In this case the page that will receive the user from the authenticator step (the "Your applications" page) will already be loading the fund + round info as well as a list of their applications, it might be better placed to make the decision to go straight into checking eligibility.

I don't think this has to be a blocker as this logic looks like its been well covered and I know we need this to get through. We could always tweak and move that logic after this pr if you think its worthwhile - happy to chat about it!

sfount (commented):
Code looks good to me! A few questions about where it lives and maybe one or things here that are unused. Happy to chat about the structure or look to get it approved and then make any changes

authenticator/frontend/templates/authenticator/partials/eligibility-check-button.html
hamzabinkhalid: good spot, included it first then removed it from the landing page, forgot to remove the references to it, thanks

authenticator/frontend/magic_links/routes.py
hamzabinkhalid: Yes, this is a valid concen, added a validaiton check in that function to check if we have a valid UUID account_id

app.py
```diff
@@ -255,6 +255,7 @@ def create_app() -> Flask:  # noqa: C901
     csrf.exempt(fund_store_bp)
     csrf.exempt(application_store_bp)
     csrf.exempt(assessment_store_bp)
+    csrf.exempt(eligibility_bp)
```
samuelhwilliams, line 258: Does this need to be exempt? If it's serving HTML and/or used by browsers, it should definitely not be - only the internal machine-to-machine 'API' endpoints should be

config/envs/development.py
```diff
@@ -57,7 +57,7 @@ class DevelopmentConfig(Config):
     USE_LOCAL_DATA = strtobool(getenv("USE_LOCAL_DATA", "True"))
     SESSION_COOKIE_SECURE = False
     AUTO_REDIRECT_LOGIN = True
-    DISABLE_NOTIFICATION_SERVICE = False  # Toggle on if you have no notify api key.
+    DISABLE_NOTIFICATION_SERVICE = True  # Toggle on if you have no notify api key.
```
samuelhwilliams, line 60: I'm not sure that this change should be committed.

If we do want to change it, maybe we should make it at least configurable eg `bool(strtobool(getenv("DISABLE_NOTIFICATION_SERVICE", "false")))`

authenticator/api/magic_links/routes.py
hamzabinkhalid: created a ticket to address this in future: https://mhclgdigital.atlassian.net/browse/FLS-911

apply/templates/apply/partials/eligibility-check-button.html
```diff
@@ -8,8 +8,7 @@
         <input type="hidden" name="round_id" value="{{ round_id }}">
 
         {{ govukButton({
-            'text': gettext('Check eligibility'),
-            'classes': 'govuk-button--secondary'
```
sfount, line 12: Just confirming that this should no longer be a secondary button? If I remember correctly the non-eligibility button is secondary will that now give them different styles?

authenticator/frontend/templates/authenticator/magic_links/landing.html
```diff
@@ -21,11 +21,12 @@ <h1 class="govuk-heading-xl">
...
 
     <div data-qa="claim_magic_link">
-    {{ govukButton({
-      "text":gettext("Continue"),
-      "href": url_for('api_magic_links.use', link_id=link_id, fund=fund_short_name, round=round_short_name),
-      "isStartButton": true })
-    }}
+        {{ govukButton({
+        "text": gettext("Continue") if not has_eligibility else
+        (gettext("Continue") if has_previous_applications else gettext("Start a new application")),
```
sfount, line 26: Maybe a bit easier to read but not too important!
```suggestion
        "text": gettext("Continue") if (not has_eligibility or has_previous_application) else gettext("Start a new application"),
```

sfount (approved):
I think we can get this in and through as it standard, we might look at where some of the responsibility for this functionality lives in a separate PR or as we refactor but this meets the needs of the ticket well, great work.

apply/templates/apply/partials/eligibility-check-button.html
hamzabinkhalid: This was a requirement to make the Eligibility button Green, would have to address the inconsistency in colors of start application in future

sfount (approved):

PR #132, opened 2025-01-08
Add Apply Front-end

albertkol:
Few things added here:
1. Banner to let the user know changes to applications were requested on the application page
2. Display CHANGE_REQUESTED status on the applications page
3. Display CHANGE_REQUESTED status to the application on the application page
4. Display CHANGE_REQUESTED status on the application page for each section
5. On re-submission of the form, mark Application and Forms as COMPLETE
6. On Form update without completion keep status CHANGE_REQUESTED


application_store/db/queries/statuses/queries.py
```diff
@@ -131,7 +135,11 @@ def update_form_status(
                 form_to_update.status = "COMPLETED"
                 form_to_update.has_completed = True
             else:
-                form_to_update.status = "IN_PROGRESS"
+                form_to_update.status = (
```
ksp37-dluhc, line 138: Not related to your change, but the conditional 
` elif "NOT_STARTED" not in status_list and is_summary_page_submitted:
`
makes me a bit nervous if we're setting the question status to CHANGE_REQUESTED. I think either we don't bother changing the question status, or we change this to something like 
` elif all( status == "COMPLETED" for status in status_list) and is_summary_page_submitted:
`

Happy to make the change in my PR though.

ksp37-dluhc (approved):

application_store/db/queries/statuses/queries.py
albertkol: Happy for you to make the change as part of your PR. 👍

PR #134, opened 2025-01-08
FSPT-208 Script to make nested JSON date strings consistent

sfount:
For a ~2 week period application `date_submitted` properties were incorrectly set with timezone information (in this case consistently UTC with `+00.00`).

The consumer of these dates (the assess tool) has been updated in places to parse any valid ISO format but leaving different formats of dates could lead to unpredictable behaviour.

Update existing assessment records to make this consistent.

Depends on https://github.com/communitiesuk/funding-service-pre-award/pull/126


samuelhwilliams:
Do we want to merge this for posterity because we maybe think it somehow secretly got run, or just close it because ... no-one remembers running it and the problem disappeared? 👀 🫣 🙈

PR #136, opened 2025-01-09
Assess frontend hotfix

albertkol:
- Reverting removing the `flags_list = get_flags(application_id)` form the applications() view. I still think it does nothing, but it's the only change I did directly inside of the view
- Removing an extra `not` form `get_change_requests` function


albertkol:
Not needed.

PR #140, opened 2025-01-09
Change format style of log messages

samuelhwilliams:
### Change description
The {} with `extras` style is used by a custom log formatted that we have in fsd-utils, which interpolates messages eagerly as they get handled.

This style is supported natively by Python logging, which means that it's better supported by things that integrate with Python - like Sentry.

Currently in Sentry, our log messages don't show up interpolated, so you'll get a raw string like "I'm a log message for {data}". Which isn't very useful if you care about the value of `data`.

By using "I'm a log message for %(data)s" instead, Sentry will still group all error logs under a single issue, but it'll also correctly interpolate the value of `data` into the message, which should aid debugging.

### DIFF
--- a/assess/scoring/routes.py
+++ b/assess/scoring/routes.py
@@ -49,7 +49,7 @@ def score(
     is_rescore = rescore_form.validate_on_submit()
     if not is_rescore and request.method == "POST":
         if score_form.validate_on_submit():
-            current_app.logger.info("Processing POST to {request_path}.", extra=dict(rqeuest_path=request.path))
+            current_app.logger.info("Processing POST to %(request_path)s.", dict(request_path=request.path))
             score = int(score_form.score.data)
             user_id = g.account_id
             justification = score_form.justification.data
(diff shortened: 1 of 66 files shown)


gidsg:
I guess the SonarCube warnings can be ignored as the logger wouldn't run code?

samuelhwilliams:
I suspect these warnings were added to sonarqube after the log4j vulnerability - it's maybe a legitimate vulnerability path. But maybe one I'd outscope for this PR..? 😬

assess/scoring/routes.py
```diff
@@ -49,7 +49,7 @@ def score(
     is_rescore = rescore_form.validate_on_submit()
     if not is_rescore and request.method == "POST":
         if score_form.validate_on_submit():
-            current_app.logger.info("Processing POST to {request_path}.", extra=dict(rqeuest_path=request.path))
+            current_app.logger.info("Processing POST to %(request_path)s.", dict(rqeuest_path=request.path))
```
srh-sloan, line 52: Typo in parameter name, should be
`current_app.logger.info("Processing POST to %(request_path)s.", dict(request_path=request.path))`

srh-sloan (changes requested):
Pulled locally, ran the python e2e tests and found one bug, otherwise all good

assess/scoring/routes.py
samuelhwilliams: good spot - thanks 🙏

srh-sloan (approved):

PR #143, opened 2025-01-09
Update Assess landing page

sc15zs:
Tickets:
https://mhclgdigital.atlassian.net/browse/FLS-891
https://mhclgdigital.atlassian.net/browse/FLS-892
https://mhclgdigital.atlassian.net/browse/FLS-894

Description:
- Changes the background colour to white.
- Updates templates to use Nunjucks syntax in header and footer

### DIFF
--- a/assess/templates/assess/footer.html
+++ b/assess/templates/assess/footer.html
@@ -1,0 +1,3 @@
+<div class="assess-footer">
+    {{ govukFooter({}) }}
+</div>
--- a/assess/static/src/styles/govuk-overrides.css
+++ b/assess/static/src/styles/govuk-overrides.css
@@ -51,7 +51,7 @@
 }
 
 .govuk-template__body {
-    background-color: #f1f1f1;
+    background-color: white;
 }
 
 .govuk-phase-banner__content {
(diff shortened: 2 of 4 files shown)


assess/templates/assess/footer.html
```diff
@@ -0,0 +1,3 @@
+<div class="assess-footer">
```
ksp37-dluhc, line 1: Is this used anywhere currently?

ksp37-dluhc (approved):

assess/templates/assess/footer.html
sc15zs: @ksp37-dluhc 
I've added this class and a css rule for this class to assess/static/src/styles/landing.css, because otherwise the footer was not always displayed on the bottom of the page. 
e.g. on the /cookie_policy page, where the main content is displayed in a 2/3-width column, and the banner was next to it, in the remaining 1/3 of screen space. And its background-colour was overlapping the other 2/3 of the screen. 
That's the reason for adding the class and the css rule, to make sure that the footer always shows on the bottom of the screen.

PR #148, opened 2025-01-10
hsra content fix

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-900


apply/templates/apply/fund_start_page.html
```diff
@@ -85,7 +85,7 @@ <h2 class="govuk-heading-m">{% trans %}Get help{% endtrans %} </h2>
                 </p>
                 {% endif %}
                 <p class="govuk-body">
-                    {% trans %}Do not send any applications or attachments from the service desk portal. We'll only accept applications through this service.{% endtrans %}
+                    {% trans %}Do not send any applications or attachments through the support desk. We'll only accept applications through this service.{% endtrans %}
```
sfount, line 88: All of these will change the translation "key" which will mean these snippets no longer load in Welsh, is that something we're OK with in the meantime or do we want to run the translate script and make sure that stays up to date?

(If these don't actually have Welsh translations at the moment then that would prove it doesn't matter?)

assessment_store/tasks/db_tasks.py
```diff
@@ -183,11 +183,40 @@ def seed_assessment_store_db(c, environment="local"):
...
                 _echo_print("No new scoring systems were added.")
 
-            # Associate scoring systems with all rounds
+            # Associate scoring systems with rounds
             round_ids = db.session.query(Round).with_entities(Round.id).all()
+            default_scoring_system_id = scoring_system_data[0]["id"]
+
+            specific_round_scoring = [
+                {
+                    "round_id": "ae223686-cbcc-4548-8b52-05898c315a59",
```
sfount, line 192: do we want fund specific logic in the code here? Something about doing this here doesn't feel right but I'll have to have a look around to see where its otherwise configured. 

Looks like everything previously defaulted to 1 to 5 and now we want to be able to override

I'd be inclined to try and put this information here or something similar https://github.com/communitiesuk/funding-service-pre-award/blob/main/assessment_store/config/mappings/assessment_mapping_fund_round.py#L231-L240

Having it spread out between configuration and tasks code feels like an (even more slippery) slope

sfount (commented):
Changes look like they up with the requirements, great work. A few hopefully quick questions before bringing it in

apply/templates/apply/fund_start_page.html
sfount: Fine with me - can we make sure following up on that gets put in as a ticket on a backlog that will be played? Happy to do that for you if you helpful

sfount (dismissed):
Approving these config changes, discussed with @hamzabinkhalid that we'll look for a more sustainable method of configuring default scoring systems and we can update pre-prod databases for demoing in the meantime

assessment_store/tasks/db_tasks.py
sfount: https://github.com/communitiesuk/funding-service-pre-award/pull/148#pullrequestreview-2547565693

hamzabinkhalid (approved):

PR #150, opened 2025-01-13
Rename ONS code to LA name script

albertkol:
Script that maps ONS code to LA name.

This will update the following:
- applications.project_title
- applications.json.category[].fields[].answer
- assessment_record.project_title
- assessment_record.jsonb_blob.project_title
- assessment_record.jsonb_blob.question[].fields[].answer


fund_store/scripts/data_updates/FLS-907-update-la-names.py
```diff
@@ -0,0 +1,525 @@
...
+                f"{json_project_name}",
+            )
+            if json_project_name in CODE_TO_NAME_MAP:
+                print(
+                    f"AssessmentRecord.json_project_name: {json_project_name} ",
+                    f"replacing with: {CODE_TO_NAME_MAP[json_project_name]}",
+                )
+                new_json_2["project_name"] = CODE_TO_NAME_MAP[json_project_name]
+            else:
+                print(f"AssessmentRecord.json_project_name not found: {json_project_name}")
```
sfount, line 481: Are the project names not being found in either of the database row or the JSON blob valid states for assessment? I would think both are necessarily there and this could be replaced by just updating the project name value on the model + the JSON 

Not a problem to be verbose in a script like this but I'd think these if conditions are almost a reason to halt the script rather than be managed like this

fund_store/scripts/data_updates/FLS-907-update-la-names.py
```diff
@@ -0,0 +1,525 @@
...
+    if commit:
+        db.session.commit()
+
+
+def main():
+    update_local_authority_name(commit=False)
+
+    # ask for confirmation
+    if input("Commit changes? (y/n): ").lower() == "y":
+        update_local_authority_name(commit=True)
```
sfount, line 516: I assume the db session changes from the first run of the `update_local_authority_name` method will be discarded or overwritten? a bit strange to run it twice but I can't think of anything it would do wrong

sfount (commented):
As a one off script this looks sensible to me and looks to do what's in the description/ requirements.

Happy to approve and merge when its ready and join on running it

fund_store/scripts/data_updates/FLS-907-update-la-names.py
albertkol: It was a simple way to get the `print()` to be echo'd 1st, so I can inspect what changes would be made. Then only on confirm have them executed. In hind-sight, it could perhaps just execute the commit, but re-printing everything is no the worst.

fund_store/scripts/data_updates/FLS-907-update-la-names.py
albertkol: No need to halt it.

In a world were we re-run the script, the already updated ONS Codes would be mentioned, but not updated.
Also, while the component is a drop-down, null or any string can be potentially stored. In this case, I am happy for it to be echo'd, but would not want to stop the sprint.

sfount (approved):
I have run this locally and it updates the data as expected - we'll try this in pre-prod environments before applying to prod

PR #167, opened 2025-01-17
Config feedback feature for GBRF/LPDF

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-929

Fixing feedback feature for GBRF/LPDF funds.


albertkol (approved):
👍 Tested and it worked.

PR #169, opened 2025-01-17
Header on Assess pages

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net/issues/FLS-920

Description:
Ensure to display header on Assess pages.


albertkol (dismissed):
Needs a big rebase, but it does the job.

I am torn about having the masthead (the big, blue element) inside the header, but I will give it a thought when we are looking into updating the overall template layout.

albertkol (approved):
:+1:

PR #171, opened 2025-01-18
Forward change requests

albertkol:
Build the dictionary of RAISED change requests and pass it to the `form-builder`. 

Structure is:
```python
{
    "question_id": ["message_1", "message_2"],
    "question_id_2": ["message_3"],
}
```

### DIFF
--- a/pre_award/apply/helpers.py
+++ b/pre_award/apply/helpers.py
@@ -79,6 +79,7 @@ def format_rehydrate_payload(
     round_name=None,
     has_eligibility=False,
     round_close_notification_url=None,
+    change_requests=None,
 ):
     """
     Returns information in a JSON format that provides the
@@ -139,6 +140,8 @@ def format_rehydrate_payload(
     formatted_data["metadata"]["fund_name"] = fund_name
     formatted_data["metadata"]["round_name"] = round_name
     formatted_data["metadata"]["has_eligibility"] = has_eligibility
+    formatted_data["metadata"]["change_requests"] = change_requests
+
     if round_close_notification_url is not None:
         formatted_data["metadata"]["round_close_notification_url"] = round_close_notification_url
 
--- a/pre_award/assessment_store/db/queries/flags/queries.py
+++ b/pre_award/assessment_store/db/queries/flags/queries.py
@@ -109,3 +109,29 @@ def resolve_open_change_requests_for_sub_criteria(
     db.session.commit()
 
     return open_change_requests
+
+
+def get_change_requests(application_id: str) -> dict[str, list] | None:
+    assessment_flags = (
+        db.session.query(AssessmentFlag)
+        .filter(
+            AssessmentFlag.application_id == application_id,
+            AssessmentFlag.latest_status == FlagStatus.RAISED,
+        )
+        .all()
+    )
+
+    if not assessment_flags:
+        return None
+
+    assessor_change_requests: dict[str, list] = {}
+    for change_request in assessment_flags:
+        for field_id in change_request.field_ids:
+            if field_id not in assessor_change_requests:
+                assessor_change_requests[field_id] = []
+
+            assessor_change_requests[field_id].extend(
+                [update.justification for update in change_request.updates if update.status != FlagStatus.RESOLVED]
+            )
+
+    return assessor_change_requests
--- a/pre_award/apply/default/application_routes.py
+++ b/pre_award/apply/default/application_routes.py
@@ -44,6 +44,7 @@ from pre_award.apply.helpers import (
     get_token_to_return_to_application,
 )
 from pre_award.apply.models.statuses import get_formatted
+from pre_award.assessment_store.db.queries.flags.queries import get_change_requests
 from pre_award.common.blueprints import Blueprint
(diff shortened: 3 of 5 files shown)


apply/helpers.py
```diff
@@ -68,6 +72,33 @@ def extract_subset_of_data_from_application(application_data: dict, data_subset_
...
+                FlagUpdate.status == FlagStatus.RAISED,
+            )
+        )
+        .all()
+    )
+
+    if not assessment_flags:
+        return {}
+
+    assessor_change_requests: dict[str, list] = {}
```
ksp37-dluhc, line 91: Could use defaultdict to save the empty list assignment below, but I'm not precious about this...

ksp37-dluhc (approved):

apply/helpers.py
```diff
@@ -68,6 +72,33 @@ def extract_subset_of_data_from_application(application_data: dict, data_subset_
     return application_data[data_subset_name]
 
 
+def get_change_requests(application_id: str) -> dict[str, list]:
```
ksp37-dluhc, line 75: This might be useful in a few places ... wonder if it might be better to have this in one of the db query files (e.g assessment_store/db/queries/flags/queries.py )

ksp37-dluhc (approved):

apply/helpers.py
```diff
@@ -68,6 +72,33 @@ def extract_subset_of_data_from_application(application_data: dict, data_subset_
...
+        for field_id in change_request.field_ids:
+            if field_id not in assessor_change_requests:
+                assessor_change_requests[field_id] = []
+
+            assessor_change_requests[field_id].extend([update.justification for update in change_request.updates])
+
+    return assessor_change_requests
+
+
 def format_rehydrate_payload(
```
sfount, line 102: This method currently receives a bunch of stuff which has been fetched from backends alreday and describes itself as formatting it in a way that can be consumed by the form runner. 

Extending it to also fetch data will work fine but might make understanding/ testing it a bit confusing if you're getting some of the data outside, so of the data in here etc.

I don't think this is a blocker as I appreciate the route handlers that call this method are hundreds of lines long already, just conscious of spreading future us knowing where to look and what patterns to follow

sfount (commented):
Nice this looks like a clean self contained change - my comments are all optional and don't need to block progress

apply/helpers.py
albertkol: Changed!

apply/helpers.py
albertkol: Moved!

apply/helpers.py
albertkol: Passing change_requests as a parameter to the function. Good call.

apply/helpers.py
```diff
@@ -68,6 +72,33 @@ def extract_subset_of_data_from_application(application_data: dict, data_subset_
     return application_data[data_subset_name]
 
 
+def get_change_requests(application_id: str) -> dict[str, list]:
+    assessment_flags = (
+        db.session.query(AssessmentFlag)
+        .join(FlagUpdate)
+        .filter(
+            and_(
```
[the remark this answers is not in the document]
albertkol: Fixed!

pre_award/apply/helpers.py
```diff
@@ -121,8 +123,9 @@ def format_rehydrate_payload(
             "constructing session rehydration payload for application id:%(application_id)s, "
             "markAsCompleteEnabled: %(markAsCompleteEnabled)s."
         ),
-        dict(application_id=application_id, markAsCompleteEnabled=markAsCompleteEnabled),
+        dict(application_id=application.id, markAsCompleteEnabled=markAsCompleteEnabled),
     )
+    application = get_application_data(application.id)
```
ksp37-dluhc, line 128: Why do we need this? Aren't we already passing in the application?

pre_award/assessment_store/db/queries/flags/queries.py
```diff
@@ -109,3 +109,30 @@ def resolve_open_change_requests_for_sub_criteria(
     db.session.commit()
 
     return open_change_requests
+
+
+def get_change_requests(application_id: str) -> dict[str, list] | None:
+    assessment_flags = (
```
ksp37-dluhc, line 115: Is the intention to get all change requests here? Or only open change requests? 

I think all assessment flags will have an associated `FlagUpdate` with FlagStatus.RAISED, so isn't the join redundant?

pre_award/apply/helpers.py
albertkol: We were passing the application_id with some form information.

I need the whole application so I can determine if it's a re-submission or not. My current form runner PR is not making use of this property yet, but it will be useful soon.

pre_award/assessment_store/db/queries/flags/queries.py
albertkol: Yep.. i switched the filter to check for the assessment_flag.latest_status instead.

ksp37-dluhc (approved):

ksp37-dluhc (dismissed):

ksp37-dluhc (approved):

PR #172, opened 2025-01-20
Add a single bootstrap command for docker compose

samuelhwilliams:
This is a short-term patch to improve startup speed for the pre-award service when running locally.

At the moment the docker-compose service does all of these things as separate commands:

```
flask -A app:create_app db upgrade
python build.py
python -m fund_store.scripts.load_all_fund_rounds
python -m fund_store.scripts.fund_round_loaders.load_fund_round_from_fab --seed_all_funds True
python -m invoke assessment.seed-assessment-store-db
python -m invoke account.seed-local-account-store
```

This results in a lot of overhead with app startup, so collapsing 5 of these into one command should mean we get to running the actual server faster.

This is low hanging fruit and I've avoided thinking about refactoring/etc

### DIFF
--- a/common/tasks.py
+++ b/common/tasks.py
@@ -1,8 +1,17 @@
 import glob
 import os
 import re
+from contextlib import contextmanager
 
+from alembic import command
+from alembic.config import Config
 from invoke import task
+
+from account_store.tasks import seed_local_account_store_impl
+from app import create_app
+from assessment_store.tasks.db_tasks import seed_assessment_store_db_impl
+from fund_store.scripts.fund_round_loaders.load_fund_round_from_fab import load_fund_from_fab_impl
+from fund_store.scripts.load_all_fund_rounds import load_all_fund_rounds
 
 _VALID_JINJA_EXTENSIONS = (".html", ".jinja", ".jinja2", ".j2")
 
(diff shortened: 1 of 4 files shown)


hamzabinkhalid (dismissed):

common/tasks.py
```diff
@@ -107,3 +116,24 @@ def pybabel_update(c):
...
+        os.environ[key] = old_val
+
+    with _env_var("FLASK_ENV", "development"):
+        with create_app().app_context():
+            alembic_cfg = Config("db/migrations/alembic.ini")
+            alembic_cfg.set_main_option("script_location", "db/migrations")
+            command.upgrade(alembic_cfg, "head")
+
+            load_all_fund_rounds()
+            load_fund_from_fab_impl()
```
srh-sloan, line 137: Should this pass `seed_all_funds=True` to the fab function?

srh-sloan (commented):
Nice job, think you've missed a param though?

common/tasks.py
samuelhwilliams: yes dang, good spot - thanks 🙏

srh-sloan (approved):

PR #176, opened 2025-01-22
Basic template loading with latest govuk-frontend

srh-sloan:
### Change description
This PR creates new base templates for Apply. This includes the standard govuk header and footer, and our cookies message.

It uses govuk frontend css and js v4.7.0

The new template is not yet accessible on any route (but I've tested it locally and on dev with a temporary route)


frontend/templates/apply/apply-base.html.jinja
```diff
@@ -0,0 +1,86 @@
+{% extends "fs-base.html.jinja" %}
+
+<meta name="description" content="Apply for a Grant">
+<meta name="keywords" content="Apply for a Grant">
+<meta name="author" content="Ministry of Housing, Communities and Local Government">
```
samuelhwilliams, line 5: I don't think these will work - when extending templates, I think everything has to be inside blocks or they won't come out in the final HTML?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "flask-talisman==1.1.0",
     "flask-wtf==1.2.2",
     "flask==3.0.3",
-    "govuk-frontend-jinja==2.7.0",
+    "govuk-frontend-jinja==3.4.0",
```
samuelhwilliams, line 32: I think this is problematic - each specific release of govuk-frontend-jinja is designed to work with a specific GOV.UK Frontend. It feels like this would probably break the existing apply/assess stuff and we should do some heavy manual QA/checks. Unless you've already done that?

frontend/apply/routes.py
```diff
@@ -0,0 +1,10 @@
+from flask import render_template
+
+from pre_award.common.blueprints import Blueprint
+
+apply_bp = Blueprint("apply_routes", __name__, template_folder="templates")
+
+
+@apply_bp.route("/monolith")
+def landing_page():
+    return render_template("apply/landing-mono.html.jinja")
```
samuelhwilliams, line 10: Do we actually want this to be a routable thing on the live deployed service?

srh-sloan: Probs not!

frontend/templates/apply/apply-base.html.jinja
srh-sloan: Moved them into a block

pyproject.toml
srh-sloan: Rolled back

build.py
```diff
@@ -193,6 +192,79 @@ def build_assess_authenticator_assets(remove_existing=False):
     static_assets.build_bundles(static_folder="static")
 
 
+def build_monolith_assets(static_dist_root="static/monolith", remove_existing=False) -> None:
+    MONOLITH_DIST_PATH = "./" + static_dist_root
+    MAIN_DIST_PATH = "./static"
+    GOVUK_URL = "https://github.com/alphagov/govuk-frontend/releases/download/v5.8.0/release-v5.8.0.zip"
```
samuelhwilliams, line 198: this is still 5.8.0 - I think if we're rolling back the python libs, we should keep using a compatible version of govuk frontend? is this too controversial?

frontend/templates/apply/apply-base.html.jinja
```diff
@@ -0,0 +1,86 @@
...
+{% block main %}
+  {{ super() }}
+{% endblock main %}
+
+{% block beforeContent %}
+  {{ govukPhaseBanner({
+      "tag": {
+        "text": "Beta"
+      },
+      "html": 'This is a new service.'
```
samuelhwilliams, line 51: Should we have a feedback/support link in here?

frontend/templates/apply/apply-base.html.jinja
```diff
@@ -0,0 +1,86 @@
...
+    })
+  }}
+{% endblock beforeContent %}
+
+{% include 'fs-cookies.html.jinja' %}
```
samuelhwilliams, line 56: I think this file including two of the blocks makes this quite hard to follow - would probably rather we have the blocks defined here and inside those include `fs-cookies-start` and `fs-cookies-end` maybe?

srh-sloan:
> could we stick to the .html file extension for now, or is .html.jinja better for you for some reason (it's microscopically worse for me 😅)?

It just means that the better-jinja extension automatically picks up these files in vs code and syntax highlights them... not previous, there's probably some other way of doing it...

frontend/templates/apply/apply-base.html.jinja
srh-sloan: I was going with matching what we already have

build.py
srh-sloan: gaaaah I thought I removed that change all together!!

frontend/apply/routes.py
```diff
@@ -0,0 +1,13 @@
...
+from pre_award.config import Config
+
+apply_bp = Blueprint("apply_routes", __name__, template_folder="templates")
+
+
+if Config.FLASK_ENV in ["dev", "development"]:
+
+    @apply_bp.route("/monolith")
+    def landing_page():
+        return render_template("apply/landing-mono.html.jinja")
```
samuelhwilliams, line 13: Ah sorry I wasn't clear :( I really really would like us to avoid any logic in our new "good" code that does checking like this on which environment we're running in to gate code/logic/features. I think we've had a few foot guns in the past with features that are meant to be developer-only ending up causing problems. This is unlikely to cause problems - of course - but I think I'd like it to be a red line for me 😓 

I think I mostly just meant: should we remove this altogether 🙏

frontend/templates/apply/apply-base.html.jinja
```diff
@@ -0,0 +1,86 @@
...
+
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/cookie-banner/macro.html" import govukCookieBanner %}
+{% from "govuk_frontend_jinja/components/skip-link/macro.html" import govukSkipLink %}
+{% from "govuk_frontend_jinja/components/header/macro.html" import govukHeader %}
+{% from "govuk_frontend_jinja/components/phase-banner/macro.html" import govukPhaseBanner %}
+{% from "govuk_frontend_jinja/components/footer/macro.html" import govukFooter %}
+
+{# TODO sort out service name, use existing mechanism for now #}
+{% set serviceName = gettext("Access Funding") if is_error else service_title or get_service_title() %}
```
samuelhwilliams, line 12: I might say that this is a thing we shouldn't carry over into the new world - can we make a simpler choice here from the start? This could be one of three things and makes it hard to track down where the service title is coming from.

If we don't have enough info to make that decision right now, that's maybe ok - but it feels like a bit of complexity we should avoid carrying over if at all possible. We'll probably need to do lots of little tidy-ups like this as we go.

frontend/templates/fs-base.html.jinja
```diff
@@ -0,0 +1,20 @@
+{% extends "govuk_frontend_jinja/template.html" %}
+
+
+{% block head %}
+  <link rel="stylesheet" href="{{ url_for('static', filename='apply/govuk-frontend-4.0.0.min.css' ) }}"> </link>
```
samuelhwilliams, line 5: If this is the full fs base template, it feels maybe weird to point to the apply govuk-frontend?

Can we do something bold here and just have a single base template rather than a fs-base and an apply-base? Ideally one base means more consistency, but depends what our thoughts are about how we do more vertical slices.

frontend/templates/apply/apply-base.html.jinja
srh-sloan: I don't think we've had a decision yet from content, but I've chased, so I was keeping it as-is for now

srh-sloan:
> > It just means that the better-jinja extension automatically picks up these files in vs code and syntax highlights them... not previous, there's probably some other way of doing it...
> 
> Does just .jinja work and is that controversial? 👀

Yes just `.jinja` seems to work equally well

frontend/templates/apply/apply-base.html.jinja
srh-sloan: Split cookies template into start and end, but include them in the fs-base file

frontend/apply/routes.py
srh-sloan: Remove this route

frontend/templates/fs-base.html.jinja
srh-sloan: Upgraded to use v4.7.0 of the frontend css and js, which is compatible with the version of the macros we already used (2.7.0)

See https://github.com/LandRegistry/govuk-frontend-jinja

frontend/templates/apply/apply-base.html.jinja
srh-sloan: Just using `get_service_title()` so we can delay the decision but get rid of the ugly stuff

frontend/apply/routes.py
srh-sloan: Gone

samuelhwilliams (dismissed):

common/templates/partials/fs-cookies-start.html
```diff
@@ -0,0 +1,89 @@
+
+{% set acceptHtml %}
+<p class="govuk-body">{% trans %}You've accepted analytics cookies. You can{% endtrans %} <a class="govuk-link"
+        href="{{ url_for('content_routes.cookie_policy') }}">{% trans %}change your cookie settings{% endtrans %}</a> {%
+    trans %}at any time.{% endtrans %}</p>
+{% endset %}
+
+{% set denyHtml %}
+<p class="govuk-body">{% trans %}You've rejected analytics cookies. You can{% endtrans %} <a class="govuk-link"
+        href="{{ url_for('content_routes.cookie_policy') }}">{% trans %}change your cookie settings{% endtrans %}</a> {%
```
sfount, line 10: How this HTML is laid out is a little chaotic for me - I don't want this to block anything but whatever is formatting this and deciding to break {% %} on to new new lines needs to stop?

samuelhwilliams (approved):

sfount (approved):
lines up with the structure we talked about! I think it will help me to see it used/ come together in order to properly review in context

common/templates/partials/fs-cookies-start.html
samuelhwilliams: agree but until we find a functional auto-formatter for jinja then I think we just have to suck this up a bit.

PR #181, opened 2025-01-23
Fix sqlalchemy/marshmallow warning on relationship

samuelhwilliams:
We've been getting this warning on app startup for a while now:

`/app/.venv/lib/python3.10/site-packages/marshmallow_sqlalchemy/convert.py:124: SAWarning: relationship 'AssessmentRecord.change_requests' will copy column assessment_records.application_id to column assessment_flag.application_id, which conflicts with relationship(s): 'AssessmentRecord.flags' (copies assessment_records.application_id to assessment_flag.application_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="flags"' to the 'AssessmentRecord.change_requests' relationship. (Background on this warning at: https://sqlalche.me/e/20/qzyx) (This warning originated from the configure_mappers() process, which was invoked automatically in response to a user-initiated operation.)`

This is because we added a (subset) relationship to AssessmentRecord. We originally had a `flags` relationship to AssessmentFlags, and we added a partial filtered relationship to the same table called `change_requests`.

Having two relationships on the same FK mean that sqlalchemy doesn't know which one to use when writing to the related table, which could lead to unexpected behaviour.

By setting the `change_requests` relationship as read-only, we can remove the warning.

### DIFF
--- a/pre_award/assessment_store/db/models/assessment_record/assessment_records.py
+++ b/pre_award/assessment_store/db/models/assessment_record/assessment_records.py
@@ -71,6 +71,7 @@ class AssessmentRecord(BaseModel):
         "AssessmentFlag",
         primaryjoin="and_(AssessmentFlag.application_id == AssessmentRecord.application_id, "
         "AssessmentFlag.is_change_request == True)",
+        viewonly=True,  # this is a subset relationship; the unfiltered `flags` should be used for writes/deletes
     )
     # These are defined as column_properties not as hybrid_property due to performance
     # Using column_property below forces the json parsing to be done on the DB side which is quicker than in python


pre_award/assessment_store/db/models/assessment_record/assessment_records.py
```diff
@@ -71,6 +71,7 @@ class AssessmentRecord(BaseModel):
         "AssessmentFlag",
         primaryjoin="and_(AssessmentFlag.application_id == AssessmentRecord.application_id, "
         "AssessmentFlag.is_change_request == True)",
+        viewonly=True,  # this is a subset relationship; the unfiltered `flags` should be used for writes/deletes
```
albertkol, line 74: If my understanding is correct.. I would rather go with `overlaps` as it's allowing both `flags` relationship and `change_requests` relationship to update `Flag` models. They both point to the same Python model, I think it should be fine.

We could make it `read_only` to put the warning to rest. But I just find it an unexpected behaviour.

I'd even rather go for:
```python
@property
def change_requests(self):
    return self.flags.filter_by(is_change_request=True).all()
```

samuelhwilliams: I'd like to explore what happens when using that property because I suspect it will in most (all?) Cases require another trip to the DB. Would be nice to not need that.

albertkol: Yes. It would make another trip to the DB. Unless self.flag was not already used before and is loaded in the memory. 
That behaviour can be changed by loading flags together with Assessents, but then we might pull those in memory even when not needed.

I actually think `override` is the way. I am not sure I see the benefit for `viewonly`, and I can absolutely see future us trying to edit `assessment_record.change_requests` and spend 3 hours confused why it won't work. 

The `property` option is indeed less efficient, but at least it's easier to understand.

albertkol (approved):
👍 Let's ship it!

PR #186, opened 2025-01-28
Fls 936 update change request page

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-936

Updating the request change page to enable assessors to add comments to individual questions. Applicants will then be able to view change requests for each specific question.

Screenshot 2025-01-28 at 10 18 40
Screenshot 2025-01-28 at 10 18 51
Screenshot 2025-01-28 at 10 19 19


pre_award/assess/assessments/templates/assessments/request_changes.html
```diff
@@ -4,6 +4,8 @@
 {% from "govuk_frontend_jinja/components/textarea/macro.html" import govukTextarea %}
 {% from "assess/macros/sub_criteria_navbar.html" import navbar %}
 {% from "assess/macros/sub_criteria_heading.html" import sub_criteria_heading %}
+{% from "govuk_frontend_jinja/components/checkboxes/macro.html" import govukCheckboxes %}
```
albertkol, line 7: govukCheckboxes is imported twice.

pre_award/assess/assessments/templates/assessments/request_changes.html
```diff
@@ -4,6 +4,8 @@
...
+{% from "govuk_frontend_jinja/components/input/macro.html" import govukInput %}
```
albertkol, line 8: govukInput, navbar and sub_criteria_heading are imported and not needed.

pre_award/assess/assessments/templates/assessments/request_changes.html
```diff
@@ -20,15 +22,12 @@
...
 <div class="govuk-grid-row">
-    <div class="govuk-grid-column-one-third">
-        {{ navbar(application_id, sub_criteria, current_theme.id) }}
-    </div>
     <div class="theme govuk-grid-column-two-thirds">
         <div class="govuk-!-margin-bottom-8">
             <h2 class="govuk-heading-l scoring-heading">Reasons for requesting change</h2>
         </div>
+
         {% if form.errors.get("justification") or form.errors.get("field_ids") %}
```
albertkol, line 31: Only field_id errors seem to be displayed in this top html element. We only need to check for field_ids.
justification no longer exists. All justification stuff can be removed.

pre_award/assess/flagging/forms/request_changes_form.py
```diff
@@ -1,27 +1,34 @@
...
+        validators = [Length(max=Config.TEXT_AREA_INPUT_MAX_CHARACTERS)]
+        if field_id in selected_field_ids:
+            validators.append(InputRequired(message="Provide a reason for requesting changes to this application"))
+        text_area_field = TextAreaField(
+            f"Reason for {question}",
+            validators=validators,
+        )
+        setattr(RequestChangesForm, field_name, text_area_field)
+
+    return RequestChangesForm
```
albertkol, line 34: Let's return the class instance `RequestChangesForm()`, rather than the class itself. So when we use it, we can do: `build_request_changes_form(question_choices)` so one less `()`.

I think that looks less frightening.

pre_award/assess/assessments/routes.py
```diff
@@ -1311,25 +1311,30 @@ def request_changes(application_id, sub_criteria_id, theme_id):
     assessment_status = determine_assessment_status(sub_criteria.workflow_status, state.is_qa_complete)
     theme_answers_response = get_sub_criteria_theme_answers_all(application_id, theme_id)
 
-    form = RequestChangesForm(
-        question_choices=[(question["field_id"], question["question"]) for question in theme_answers_response]
-    )
+    question_choices = [
+        (question["field_id"], question["question"], question.get("answer", "")) for question in theme_answers_response
+    ]
```
albertkol, line 1316: We only really need `field_ids` to pass to the Form class. I would change this to be just an array of field_ids. We should only pass information we need to the class/function.

You are using the `question` in the Form class for the textarea description, we can rewrite that to be "Reason for {field_id}" or just "Reason".

pre_award/assess/assessments/routes.py
```diff
@@ -1311,25 +1311,30 @@ def request_changes(application_id, sub_criteria_id, theme_id):
...
 
+    form = build_request_changes_form(question_choices)()
+    justification_data = {field_id: getattr(form, f"reason_{field_id}").data for field_id, _, _ in question_choices}
```
albertkol, line 1319: You only really need the `justification_data` inside the POST if, please move it inside.

Also.. I would try to do this field_id to reason_field_id mapping on the form data submitted, not all the fields. That way you will always have have a non-None reason textareas. Saving you an if.

pre_award/assess/assessments/routes.py
```diff
@@ -1311,25 +1311,30 @@ def request_changes(application_id, sub_criteria_id, theme_id):
...
-            application_id=application_id,
-            flag_type=FlagType.RAISED.name,
-            user_id=g.account_id,
-            justification=form.justification.data,
-            field_ids=form.field_ids.data,
-            section=[sub_criteria_id],
-            is_change_request=True,
-        )
+        for field_id, justification in justification_data.items():
+            justification = justification.strip()
```
albertkol, line 1322: Good practice is to not modify objects/dictionary properties, as it can lead to unexpected behaviour later in your code. I'd create a separate variable, or in this case no bother to store it and do a quick return

You can do a quick return:
```python
if not justification.strip():
    continue
```

pre_award/assess/assessments/routes.py
```diff
@@ -1346,8 +1351,8 @@ def request_changes(application_id, sub_criteria_id, theme_id):
         "assessments/request_changes.html",
         form=form,
         question_choices=[
-            {"text": label, "value": value, "checked": value in (form.field_ids.data or [])}
-            for value, label in form.field_ids.choices
+            {"text": label, "value": value, "response": answer, "checked": value in (form.field_ids.data or [])}
```
albertkol, line 1354: I think you can use directly `theme_answers_response` to build this dictionary for the template.

The tuple we had before had too many items, that we only really needed here: question and answer property.

So when we were looping we were doing the `_`, as we were not using them. I think other programming languages don't mind it. But in Python I don't see it often. I use it only when I am lazy. But I wanted to avoid it here.

albertkol (changes requested):
Ayo!

albertkol (approved):
Pretty! Well done. 💯

PR #196, opened 2025-01-29
FSPT-240 Adding mypy

srh-sloan:
### Change description

Adding `mypy` to enforce typing across new code.
- Excludes the old `pre_award` folder
- Added to `pre-commit` so that the checking runs on a commit. This setup doesn't work with pre-commit.ci however, as the official mirror for mypy does not support all the options we are using to exclude the `pre_award` folder, and intentially ignores pyproject.toml. So I've added to this repo's pipeline instead until we have a codebase that is all mypy compliant.
- Can be run manually with `uv run mypy`
- Made the following changes to be mypy compliant:
        - Changed models to use Mapped[] instead of redefining things
        - Tidied up data types and column names (new migration 010)
        - Added return types and None checks where necessary
        - Renamed properties on fund and round to be consistent (name instead of fund_name etc)

### How to test
Pull this branch
Run `uv run mypy` and you should see n errors

Screenshot 2025-01-29 at 10 27 28


.github/workflows/copilot_deploy.yml
```diff
@@ -43,6 +43,25 @@ jobs:
...
+        with:
+          python-version: 3.10.x
+      - name: Install uv
+        uses: astral-sh/setup-uv@887a942a15af3a7626099df99e897a18d9e5ab3a # v5
+        with:
+          enable-cache: true
+      - name: install dependencies
+        run: uv sync
+      - name: run type checking
+        run: uv run mypy
```
samuelhwilliams, line 63: :chefkiss: nice - need to do this on post-award too because otherwise you have to source the venv.

.pre-commit-config.yaml
```diff
@@ -36,3 +37,12 @@ repos:
...
 #       - id: djlint-jinja
 #         types_or: ['html', 'jinja']
+
+  - repo: local
+    hooks:
+      - id: mypy
+        name: Run type checks
+        language: system
+        types: [python]
+        entry: mypy
```
samuelhwilliams, line 47: Can this run as `uv run mypy`, or does that break? 👀 

Reason being - otherwise you have to source the venv for mypy to be available, whereas `uv` should take care of that?

pyproject.toml
```diff
@@ -119,4 +119,25 @@ dev = [
...
+module = [
+    "flask_babel",
+]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = [
+    "pre_award.*"
+]
+ignore_errors = true
```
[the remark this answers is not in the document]
srh-sloan: Yeah, maybe we can talk about it at the next COP

.pre-commit-config.yaml
srh-sloan: Yeah that seems to work - done!

pre_award/db/migrations/versions/010_data_standards.py
```diff
@@ -0,0 +1,43 @@
...
+Revises: 009_relax_flag_json_constraints
+Create Date: 2025-01-29 09:50:21.026372
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "010_"
```
samuelhwilliams, line 15: might be nice for this to have a sensible revision id, although I appreciate it's _close_ to a noop

pre_award/db/__init__.py
```diff
@@ -10,8 +14,24 @@
     "pk": "pk_%(table_name)s",
 }
 
+
+class FundingType(enum.Enum):
+    COMPETITIVE = "COMPETITIVE"
+    UNCOMPETED = "UNCOMPETED"
+    EOI = "EOI"
```
samuelhwilliams, line 21: I think, as we start to add more types and have more custom defintions, we probably want a types package/module (internal to this repo)

But for now this is fine

samuelhwilliams (approved):
Nice work, this should be super useful for making sure we have a more reliable/understandable codebase in the new world

PR #198, opened 2025-01-29
Using macro with tudor crown now jinja updated

srh-sloan:
Removing TODO from the vertical slice work and using the `govukHeader` jinja macro on the new `apply-base.html`.
Also updates the version of the css and js that is brought in to the new apply to `4.8.0` as that matches the new jinja macro version.

It's only used by the landing page so doesn't need to the sign out bit yet, will add that in the future when this page is used more widely

### DIFF
--- a/apply/static/src/styles/apply.css
+++ b/apply/static/src/styles/apply.css
@@ -1,4 +1,4 @@
-/* TODO Remove once frontend jinja macros updated and we can use macros for headers - see apply-base.html */
-.header-pointer-events-none {
+/* Custom CSS for Apply */
+a.govuk-header__link--homepage {
     pointer-events: none;
   }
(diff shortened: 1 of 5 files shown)


apply/static/src/styles/apply.css
```diff
@@ -1,4 +1,4 @@
-/* TODO Remove once frontend jinja macros updated and we can use macros for headers - see apply-base.html */
-.header-pointer-events-none {
+/* Custom CSS for Apply */
+a.govuk-header__link--homepage {
     pointer-events: none;
   }
```
samuelhwilliams, line 4: Any idea why we have this custom code? Feels like we should just keep the normal behaviour unless there's a really good reason?

samuelhwilliams (approved):

apply/static/src/styles/apply.css
srh-sloan: It's so you don't get the underline hover behaviour on the gov.uk text and logo. It doesn't go anywhere so no point in making it look like a link

PR #200, opened 2025-01-29
Remove unused template

srh-sloan:
Following on from FSPT-239, this removes the old landing page template and recompiles translations to remove that page


gidsg (approved):
👍 🔴

PR #212, opened 2025-02-04
Split out CI and deployment pipelines

samuelhwilliams:
### Change description
This should do a few things:

1) Make it easier to see only the pipeline runs that are doing actual deployments to environments. Currently these are intermingled with CI test runs.

2) Disable the DB migrations check on the deployment pipeline. This check is used for CI only, because in some situations we might expect (and want) it to fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+name: Build and unit test
+
+on:
+  push:
+
+jobs:
+  setup:
+    uses: communitiesuk/funding-service-design-workflows/.github/workflows/determine-jobs.yml@main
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
Put this up as an option for unblocking @srh-sloan on https://github.com/communitiesuk/funding-service-pre-award/pull/208, which can't be deployed because we (intentionally) don't have migrations that cover all of the python model changes.

We're going to do a deploy-with-downtime to get past this and then will regroup to fix our pipeline so that we can intentionally deploy without having up-to-date migrations (we need this in cases where we're removing columns, which should be done across multiple deployments).

PR #213, opened 2025-02-04
Updating pipeline to only run checks on a ci run, not a deploy

srh-sloan:
### Change description
Updating our pipeline to skip the following checks on a deployment (but keep them on a CI run):
- type_checking
- check_db_migrations

This is to prevent the issue we saw [here](https://funding-service-design-team-dl.sentry.io/issues/6237646332/?referrer=slack&alert_rule_id=15583726&alert_type=issue) deploying database changes whereby the DB column is removed before the code is updated. Only running `check_db_migrations` on a CI run means that we can deploy column removal DB changes as follows:
1. PR to make the column nullable (with migration) and remove the column from the model (no migration, just a change in the model file). Also remove any references to that column or field.
2. PR with the migrations that remove the columns (by this point no code references that column so it won't produce errors)

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -37,14 +37,15 @@ jobs:
       environment: ${{ inputs.environment }}
 
   pre_deploy_tests:
+    needs: [setup]
     uses: communitiesuk/funding-service-design-workflows/.github/workflows/pre-deploy.yml@main
     with:
       postgres_unit_testing: true
-      check_db_migrations: true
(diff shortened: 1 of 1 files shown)


samuelhwilliams (dismissed):
Nice. Please remove the comment and add a decent PR description explaining the issue though 🙏

srh-sloan:
> Nice. Please remove the comment and add a decent PR description explaining the issue though

Done - that's why it was in draft ;-)

samuelhwilliams (approved):

PR #220, opened 2025-02-05
Fspt 246 citext

srh-sloan:
### Change description
We often need to compare the `short_name` of a fund or round with the query param value, and use `upper()` or `lower()` to do case-insensitive comparisons. This PR changes the column definition of `short_name` on `fund` and `round` to use `CITEXT` for case-insensitive comparisons at the DB level.

https://www.postgresql.org/docs/current/citext.html

### How to test
Load the apply landing page and use normal fund and round codes in different cases - should still direct you to the right fund and round

N/A


samuelhwilliams (approved):
Tested this a bit locally and all looks good to me.

There's a small nitpicky part of me that would want to redirect users to the uppercase grant code/round code URL if they enter one in lower/mixed case - but I think definitely not important here. I've run the e2e tests (python+js) and that gets through all the journeys successfully, so I think the risk of anything in the system caring about a particularly casing any more is low.

PR #228, opened 2025-02-06
Fls 957 email notification change request

ksp37-dluhc:
### Change description
This PR implements an email notification when a change request is submitted by an assessor. It will only be sent if it is the first change request of the day (to prevent spamming)

### How to test
Create a change request. The first change request should trigger an email, the second should not.

### DIFF
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -7,7 +7,7 @@ import json
 import time
 import zipfile
 from collections import OrderedDict
-from datetime import datetime
+from datetime import datetime, timezone
 from urllib.parse import quote_plus, unquote_plus
 
 from flask import Response, abort, current_app, g, redirect, render_template, request, session, url_for
@@ -114,6 +114,7 @@ from pre_award.assess.services.data_services import (
     get_tags_for_fund_round,
     get_users_for_fund,
     match_comment_to_theme,
+    notify_applicant_changes_requested,
     submit_change_request,
     submit_comment,
     update_assessment_record_status,
@@ -139,6 +140,7 @@ from pre_award.assess.themes.deprecated_theme_mapper import (
     order_entire_application_by_themes,
 )
 from pre_award.assessment_store.db.models.assessment_record.enums import Status as WorkflowStatus
+from pre_award.assessment_store.db.queries.flags.queries import is_first_change_request_for_date
 from pre_award.assessment_store.db.queries.scores.queries import approve_sub_criteria
 from pre_award.common.blueprints import Blueprint
 from pre_award.config import Config
@@ -1342,6 +1344,8 @@ def request_changes(application_id, sub_criteria_id, theme_id):
     field_ids = [question["field_id"] for question in theme_answers_response]
     form = build_request_changes_form(field_ids)
     if request.method == "POST" and form.validate_on_submit():
+        today_date = datetime.now(tz=timezone.utc).date()
+        send_notification = is_first_change_request_for_date(application_id=application_id, date=today_date)
         justification_data = {field_id: getattr(form, f"reason_{field_id}").data for field_id in field_ids}
         for field_id, justification in justification_data.items():
             if not justification.strip():
@@ -1360,6 +1364,10 @@ def request_changes(application_id, sub_criteria_id, theme_id):
                 application_id=application_id,
                 status=WorkflowStatus.CHANGE_REQUESTED,
             )
+            
+            if send_notification:
+                notify_applicant_changes_requested(application_id=application_id)
+                send_notification = False
 
         mark_application_with_requested_changes(application_id=application_id, field_ids=form.field_ids.data)
 
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -1364,7 +1364,7 @@ def request_changes(application_id, sub_criteria_id, theme_id):
                 application_id=application_id,
                 status=WorkflowStatus.CHANGE_REQUESTED,
             )
-            
+
             if send_notification:
                 notify_applicant_changes_requested(application_id=application_id)
                 send_notification = False
--- a/pre_award/assess/services/data_services.py
+++ b/pre_award/assess/services/data_services.py
@@ -7,6 +7,10 @@ import requests
 from flask import abort, current_app
 from sqlalchemy import select
 
+from data.models import Fund as FundModel
+from data.models import Round as RoundModel
+from pre_award.account_store.db.models.account import Account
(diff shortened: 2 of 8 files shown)


pre_award/config/envs/default.py
```diff
@@ -195,7 +195,7 @@ class DefaultConfig(object):
 
     # Applicant Frontend
     APPLICANT_FRONTEND_HOST = environ.get("APPLICANT_FRONTEND_HOST", "frontend")
-
+    FUND_APPLICATION_ENDPOINT = "/funding-round/{fund_short_name}/{round_short_name}"
```
gidsg, line 198: Can we keep things that don't change per environment out of the config? So move this inline?

ksp37-dluhc: Sure, I've updated it to remove it from the config.

pre_award/assessment_store/db/queries/flags/queries.py
```diff
@@ -14,14 +14,46 @@ def get_flags_for_application(application_id):
     return results
 
 
-def get_change_requests_for_application(application_id):
+def get_change_requests_for_application(application_id, only_raised=False, sort_by_update=False, sort_by_raised=False):
```
ksp37-dluhc, line 17: This is a WIP since it will need extending for other tickets. Have left a TODO for now.

ksp37-dluhc:
pre-commit.ci autofix

pre_award/assess/assessments/routes.py
```diff
@@ -1342,6 +1344,8 @@ def request_changes(application_id, sub_criteria_id, theme_id):
     field_ids = [question["field_id"] for question in theme_answers_response]
     form = build_request_changes_form(field_ids)
     if request.method == "POST" and form.validate_on_submit():
+        today_date = datetime.now(tz=timezone.utc).date()
+        send_notification = is_first_change_request_for_date(application_id=application_id, date=today_date)
```
samuelhwilliams, line 1348: I have concerns about this "only send one email a day" approach - do we have evidence that this is really required and that it won't confuse users?

I can envision a situation where an assessors asks for a change, the user gets the email and makes the change, then an hour later the assessor asks for another change. The user has no idea this has happened and thinks their application is fine, and then they move on to other things and never discover that their application needs further work.

I think - unless there's some really strong evidence here and we've thought about edge cases or repercussions - we should keep this simple and just always send an email?

Or are we expecting 10s/100s of change requests and we absolutely need to rate limit the emails somehow?

ksp37-dluhc: It's definitely something we've discussed at length in the team, and the consensus was that there's no perfect solution but this is a reasonable compromise. Like you say, there are edge cases with this as well, but the greater concern was spamming applicants every time a change request is made. I can't speak to the user research but it's a product decision rather than a development one. 

(I do agree that that there is ambiguity over how this will perform in practice, but it's easy to fallback to an email per request if we need to.)

samuelhwilliams: Cool, if product have made the call then that's all good - thanks for explaining 👍

ksp37-dluhc: I forgot to add; I don't think the situation you described should occur. The flow should be as follows:
- The assessor requests a change, which is a flag with status `RAISED`.
- The applicant makes the change, and any others. 
- The applicant resubmits the application (only possible after all changes are addressed). On re-submission, all `RAISED` flags will have their status changed to `RESOLVED`.
- The assessor requests another change (on the same day). The query will only consider flags with a status of `RAISED`, and since there are none after re-submission, an email should be sent.  

This only works because we have a re-submission process, which might not be the case in future.

samuelhwilliams: ah interesting - nice, that all makes sense then.

Is it possible to re-submit without addressing any of the change requests? Could this happen:

* Assessor makes a change request
* Applicant logs in and fixes it
* Assessor makes another change request
* Application submits the application without knowing about the 2nd change request

pre_award/assess/services/data_services.py
```diff
@@ -857,6 +862,26 @@ def get_scoring_system(round_id: str) -> List[Flag]:
     return scoring_system
 
 
+def notify_applicant_changes_requested(application_id: str):
```
sfount, line 865: This isn't a high priority and definitely doesn't need to block this PR. Having this as a data service feels a bit strange to me as it seems like it will only be called by code that already has some or most of this information in order for it to decide that you should actually send this email. Just writing this up as we'll be thinking about how we want to structure code and any discussions and opinions on it are useful to help shape that.

I wonder if the HTTP route handler that calls this could fetch the application, it could then almost directly call the notification service without separately fetching another copy of everything synchronously from the database.

How I imagine that would work
```python
@assessment_bp.route()
def the_handler():
    # ideally an application already joins account and round (which joins fund) so that information should
    # all be usable in the notification service
    application = get_application(id)

    # the account email, references, fund and round names and deadline are all then unpacked to shape the
    # personalisation over in the notification service helper
    get_notification_service().send_requested_changes_email(application=application)
```

pre_award/assess/services/data_services.py
```diff
@@ -857,6 +862,26 @@ def get_scoring_system(round_id: str) -> List[Flag]:
...
+
+    get_notification_service().send_requested_changes_email(
+        email_address=account.email,
+        application_reference=application.reference,
+        fund_name=fund.name_json[language.name],
+        round_name=round.title_json[language.name],
+        apply_fund_url=apply_fund_url,
+        application_deadline=round_deadline,
+        contact_email=round.contact_email,
+    )
```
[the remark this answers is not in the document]
ksp37-dluhc: Thanks Steven - I've wrestled with this myself a few times (what should go where, and how data should be retrieved) so I'm glad to be having this discussion! In the old world, it seems data services was responsible for retrieving data, so I continued with that pattern. (Essentially treating the DB as a service, as we did when we had separate stores).

One question we need to ask is if we feel comfortable having SQLAlchemy models at the handler level, or if they should be serialised to some other format (dict, json, dataclass etc). I think @samuelhwilliams raised this at some point. I wanted to avoid going down a path one way or another until we've made that decision. I see arguments on both sides...

There's also a painful hangover from the separate stores - we don't have referential integrity between applications, accounts, funds/rounds etc. That's why I've had to fetch them separately instead of using a join.

albertkol (approved):
👍 Tested and it works. It did not send a second email, as expected.

If others don't mind, I will approve it. I think it can go out.

ksp37-dluhc:
pre-commit.ci autofix

albertkol (approved):

PR #229, opened 2025-02-07
FSPT-228: Override frontend hostnames on Dev

gidsg:
### Change description
 Override frontend hostnames on Dev to test domain migration

### How to test
Journeys should work as before if they start at:
https://apply.dev.access-funding.communities.gov.uk/ or https://assess.dev.access-funding.communities.gov.uk/

### DIFF
--- a/copilot/fsd-pre-award/manifest.yml
+++ b/copilot/fsd-pre-award/manifest.yml
@@ -95,7 +95,7 @@ environments:
   dev:
     variables:
       ALLOW_ASSESSMENT_LOGIN_VIA_MAGIC_LINK: true
-      COOKIE_DOMAIN: ".access-funding.dev.communities.gov.uk"
+      COOKIE_DOMAIN: ".access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
       ASSESS_HOST: "assess.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
       ASSESSMENT_FRONTEND_HOST: "https://assess.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
       APPLICANT_FRONTEND_HOST: "https://apply.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
--- a/copilot/fsd-pre-award/manifest.yml
+++ b/copilot/fsd-pre-award/manifest.yml
@@ -96,8 +96,8 @@ environments:
     variables:
       ALLOW_ASSESSMENT_LOGIN_VIA_MAGIC_LINK: true
       COOKIE_DOMAIN: ".dev.access-funding.communities.gov.uk"
-      ASSESS_HOST: "https://assess.${COPILOT_ENVIRONMENT_NAME}.access-funding.communities.gov.uk"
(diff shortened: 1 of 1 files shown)


Tiny49 (approved):
LGTM

copilot/fsd-pre-award/manifest.yml
```diff
@@ -110,6 +121,20 @@ environments:
...
     http:
+      alias:
+        - account.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk
+        - assess.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk
+        - apply.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk
+      additional_rules:
+        - path: /
+          target_container: nginx
+          alias:
+            - apply.access-funding.dev.communities.gov.uk
```
sfount, line 132: Not 100% sure with this additional rules alias is doing but does this communities one need to be here? looks like the others are for aliasing the previous levellingup domains. This is also the only one that isn't relative to the `${COPILOT_ENVIRONMENT_NAME}`

gidsg: @sfount Ah, yes, this is v tricksy, it's because we hit the limit of 5 conditions on an ALB listener rule, as we need 6 aliases here, so we need to use the `additional_rules` to work around this.

This is as per the guidance from the copilot CLI:
```
You can split the "alias" and "allowed_source_ips" field into separate rules, so that each rule contains up to 5 values: 

http:
  path: /
  alias: [frontend.dev.access-funding.test.levellingup.gov.uk,authenticator.dev.access-funding.test.levellingup.gov.uk,account.access-funding.dev.communities.gov.uk,assess.access-funding.dev.communities.gov.uk]
  allowed_source_ips: []
  additional_rules:
    - path: /
      alias: [apply.access-funding.dev.communities.gov.uk]
      allowed_source_ips: []
✘ validate manifest against environment "dev": validate "http": validate condition values per listener rule: listener rule has more than five conditions 
```

See also: https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#http-additional-rules

> This is also the only one that isn't relative to the ${COPILOT_ENVIRONMENT_NAME}

Yes, good spot, I can change it to use that variable although it wouldn't make any difference yet as we're only doing it in dev so far.

gidsg: Ah also you're right, it should be levellingup.gov.uk not communities on those additional ones 🤦‍♂️

gidsg: Also good spot @sfount the `apply` one there is a typo , will fix now!

gidsg:
As commit history is messy from a lot of trial and error am thinking to squash and merge after approval, unless people have strong objections to that.

Tiny49 (approved):
LGTM

copilot/fsd-pre-award/manifest.yml
gidsg: Fixed in [3bc6e8c](https://github.com/communitiesuk/funding-service-pre-award/pull/229/commits/3bc6e8c941b7ff3eef490cb34c950f8fba493ecc)

PR #238, opened 2025-02-10
Block sections without feedback

albertkol:
This PR does two things:
- fixes a bug:
  - if you are a user that received feedback
  - and you click on the section with feedback and at the end mark it NOT complete
  - it will update the application's status as CHANGES REQUESTED
  - it should stay In Progress
- blocks all sections that did not receive CHANGE REQUESTS on a re-submission


israr-ulhaq (approved):
LGTM!

PR #244, opened 2025-02-11
Test Containers

srh-sloan:
### Change description
This change makes all the unit tests that need a database use the `testcontainers` library to spin up a postgres instance inside a container, rather than relying on the docker-compose setup locally.

It updates all existing uses of the following fixtures to just use the `db` fixture instead:
- `clear_test_data`
- `enable_preserve_test_data`
- `_db`

Testcontainers also work in the pipeline, so the workflow has been changed to pass `postgres_unit_testing: false` to the shared workflow now to avoid spinning up a container manually as we used to do.  

This also enables parallel unit testing and adds the `pytest-xdist` library to allow this. To run tests in parallel:
`pytest -n 2` where `2` is the number of parallel workers you require.

### How to test
Run unit tests locally or in the pipeline.


samuelhwilliams (approved):
This is great ✅  A few bits of cruft we can just delete, I think.

samuelhwilliams (approved):

PR #250, opened 2025-02-12
Moving unit tests into main workflow

srh-sloan:
This PR moves the pre_deploy jobs (unit tests, db migrations, type checking) out of the workflows repo and into this github action directly, and then updates the run tests command to run the tests with 2 parallel workers


.github/workflows/copilot_deploy.yml
```diff
@@ -61,7 +59,16 @@ jobs:
...
+        env:
+          FLASK_ENV: "development"
+        run: uv run python build.py
+      - name: run unit tests
+        env:
+          GOV_NOTIFY_API_KEY: ${{ secrets.GOV_NOTIFY_API_KEY }}
+        run: uv run pytest -n 2 .
       - name: run type checking
+        if: ${{ needs.setup.outputs.jobs_to_run == '[]' }}
         run: uv run mypy
```
samuelhwilliams, line 72: This will add a bit of time to the test run as this is now running sequentially rather than in parallel. Do we want to split this out?

samuelhwilliams (dismissed):

.github/workflows/copilot_deploy.yml
srh-sloan: Made all the checks run in parallel

samuelhwilliams (approved):

PR #266, opened 2025-02-14
Inline deployment pipeline

samuelhwilliams:
Pre-requisite for https://mhclgdigital.atlassian.net/browse/FSPT-212

It is probably easier to just read the files from top-to-bottom in this PR, rather than reviewing the diff.

### Change description
This patch starts by pulling in the deployment workflow job from our shared workflows repository
(https://github.com/communitiesuk/funding-service-design-workflows) before heavily modifying it to address a number of issues:

1) When deploying, the deployment and e2e tests should run as a single batched unit, so that the tests are actually validating the deployed changes. Before this patch, they would run with a break in the middle, which meant if you had multiple pipelines the following could happen:

  * deploy 1
  * deploy 2
  * tests 1
  * tests 2

The tests for deploy 1 are then not really valid.

2) The pipeline didn't nicely support deploying copilot scheduled tasks (jobs), which we now need to add.

3) Deploying to dev currently ... doesn't work (deploy job gets skipped). Could definitely fix that separately, but also - this will just fix that as a side effect.

## Examples
[Deployment to dev environment](https://github.com/communitiesuk/funding-service-pre-award/actions/runs/13324882406)
[Deployment to test environment](https://github.com/communitiesuk/funding-service-pre-award/actions/runs/13325813621)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,213 @@
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      image_location:
+        description: "Location of the image to deploy."
+        type: string
+        required: false
+      slack_alert_on_failure:
+        description: "Sends an alert to the prod alerts channel if deployment fails"
+        required: true
+        default: false
+        type: boolean
+      notify_slack_on_deployment:
+        description: "Send messages to the deployments channel when deploys start+finish."
+        default: false
+        type: boolean
+    secrets:
+      AWS_ACCOUNT:
+        required: true
+      FSD_GH_APP_ID:
+        required: true
+      FSD_GH_APP_KEY:
+        required: true
+      FS_BASIC_AUTH_USERNAME:
+        required: true
+      FS_BASIC_AUTH_PASSWORD:
+        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,213 @@
...
+
+  deploy:
+    name: ${{ matrix.deployment }}
+    needs: [ db_migrations ]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - deployment: pre-award
+            command: svc
```
samuelhwilliams, line 115: This is the bit that will let me easily add a deployment for the scheduled job

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,213 @@
...
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      image_location:
+        description: "Location of the image to deploy."
+        type: string
+        required: false
+      slack_alert_on_failure:
```
srh-sloan, line 12: Really nitpicky 🤏 but could this be called `alert_slack_on_failure` so it lines up with `notify_slack_on_deployment` below please?

srh-sloan (approved):
One tiny comment, otherwise looks good! Easier to follow 👌

srh-sloan (approved):
Nice 🎉

PR #268, opened 2025-02-14
Restore missed inputs

samuelhwilliams:
I missed these inputs, which means the pipeline template is invalid for running the tests :( 

https://github.com/communitiesuk/funding-service-pre-award/actions/runs/13326667390


albertkol (approved):
:+1:

PR #269, opened 2025-02-14
Remove some logs

albertkol:
Remove some logs


albertkol:
Maybe we shouldn't do this.

PR #271, opened 2025-02-14
Fix slack notification status for prod

samuelhwilliams:
We don't run e2e tests in prod, but that shouldn't cause the slack notification to be a failure.


albertkol (approved):
much wow

PR #276, opened 2025-02-17
Fix Application status update bug

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net/browse/FLS-1003

Description:
On Assess, adding a comment to the whole application now will not update the application status to COMPLETED.


pre_award/assess/assessments/routes.py
```diff
@@ -1537,9 +1536,6 @@ def application(application_id):
     :return:
     """
 
-    if request.method == "POST":
-        update_ar_status_to_completed(application_id)
```
hamzabinkhalid, line 1541: After scoring/accepting the changes in the subcriteria, we get the "All sections assessed" banner, clicking on "Mark as complete" should change the status to "Assessment Complete", Since we deleted the call to update_ar_status_to_completed it's not doing that right now, this seems to be done [here](https://github.com/communitiesuk/funding-service-pre-award/blob/2a6f8bf439646e6f3ab3aadc5b8acb41d92fb264/pre_award/assess/templates/assess/macros/assessment_completion.html#L7)

image

sc15zs: I've added a hidden input to the form when adding / editing a comment, and in these cases it will not call the 'update_ar_status_to_completed' function to update the Application status. In all other cases it still will. 
I had to separate this into another function, ruff didn't allow adding it to the application function, it complained about being too complex.

hamzabinkhalid (approved):
Brilliantly written test ✨

pre_award/assess/assessments/routes.py
```diff
@@ -1528,6 +1528,12 @@ def download_multiple_files(files, folder_name):
     )
 
 
+def handle_application_post_request(application_id):
+    action = request.form.get("action", None)
+    if request.method == "POST" and action != "save_comment":
```
sfount, line 1533: This doesn't have to change or block this pull request but having this check for what it's _not_ doing feels like if any other features are added or changed on this page we could end up right back at this bug because the post request will run this as long as its not about comments.

If it would explicitly look for `action == "complete_assessment"` or similar then the other features on the page changing should never mean this code could be called.

This change is just looking to patch a bug so isn't responsible for this but it also feels fragile that we'll update the status of something without doing any checks or considering its place in the workflow. It all seems well defined in the domain/ UX (i.e in order to complete an assessment you want it to be in progress, not currently flagged and have all required sections scored) but we don't reflect that in the code.

sc15zs: Yes, I agree with you, that would have been a much better solution, I was literally just stopping it to happen when adding or editing a comment. Nothing stops us updating it in the future though. :simple_smile:

PR #279, opened 2025-02-18
Update applicant export feature

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-983

This update allows to export applicant data based on both funds and rounds. If a round exists for a fund, it grabs the data for that round; if not, it falls back to the main fund data. This way, adding new rounds won’t impact the existing setup. We’ve also added a config for LPDF Round 2.


albertkol (commented):
Could you please also write some tests for `get_application_data_for_export`?

What happens when?
- there is no fund or fund:round config defined
- there is only fund config defined
- there is both fund and fund:round config defined

albertkol (approved):
👍

PR #284, opened 2025-02-20
 Rename the label Review changes in the status

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-1013

### DIFF
--- a/pre_award/assess/config/display_value_mappings.py
+++ b/pre_award/assess/config/display_value_mappings.py
@@ -30,7 +30,7 @@ assessment_statuses = {
     "STOPPED": "Stopped",
     "MULTIPLE_FLAGS": "Multiple flags to resolve",
     "CHANGE_REQUESTED": "Change requested",
-    "CHANGE_RECEIVED": "Change received",
+    "CHANGE_RECEIVED": "Review changes",
 }
 
 funding_types = {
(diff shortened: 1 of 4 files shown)


sc15zs (approved):
👍

hamzabinkhalid:
Is it only for Assess?

pre_award/assess/config/display_value_mappings.py
```diff
@@ -30,7 +30,7 @@
     "STOPPED": "Stopped",
     "MULTIPLE_FLAGS": "Multiple flags to resolve",
     "CHANGE_REQUESTED": "Change requested",
-    "CHANGE_RECEIVED": "Change received",
+    "CHANGE_RECEIVED": "Review changes",
```
hamzabinkhalid, line 33: Shouldn't we can change the key "CHANGE_RECEIVED" to "REVIEW_CHANGES"?

hamzabinkhalid:
Label in section(subcriteria) level also needs to be changed from "Change received" to "Review changes"

https://mhclgdigital.atlassian.net/browse/FLS-863

e.g. https://github.com/communitiesuk/funding-service-pre-award/blob/84e11a83533f0ed5efa0358c22fb75af0b8c5417/pre_award/assess/templates/assess/macros/criteria_element.html#L13

pre_award/assess/config/display_value_mappings.py
israr-ulhaq: Since we're already using CHANGE_RECEIVED status everywhere, it makes sense to keep it the same for the banner as well. This way, everything stays aligned with the existing enum.

sc15zs (approved):
👍

PR #293, opened 2025-02-25
Fix Apply banner notification when change requested

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-1014

This update fixes the apply banner notification. The applicant will now see the banner if any form has a change request status.

Screenshot 2025-02-25 at 11 28 31


albertkol (approved):
GG WP

PR #295, opened 2025-02-27
fix banner and add tests

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-1014

This update fixes the apply banner notification. The applicant will now see the banner if any form has a change request status.


albertkol:
Did you try deploying it to test to see if the e2e tests pass?

albertkol (approved):
:+1:

PR #304, opened 2025-03-03
Fix request change form

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-997

This will now display only the fields that have data from the applicant, enabling the assessor to request changes.

Screenshot 2025-03-03 at 12 24 37
Screenshot 2025-03-03 at 12 24 50
Screenshot 2025-03-03 at 12 25 11
Screenshot 2025-03-03 at 12 25 22


Kateharries (approved):
This looks good to me.

PR #305, opened 2025-03-03
Improve Apply section blocking after receiving change request

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net/browse/FLS-1029

Description:
In order to avoid the situation described in the ticket, after receiving a change request, we no longer use the section status to determine whether to allow editing a section or not, but we use the "field_ids" of the AssessmentFlag instances for the given application, and compare this value with the form name in the tasklist route. 


israr-ulhaq (approved):
Looks good to me!

PR #309, opened 2025-03-04
Only display uploaded file that matches answer in Assess

ksp37-dluhc:
### Change description
This partially closes off a bug that only appears in the uncompeted journey.
Previous behaviour: display all files in the s3 bucket for that field
New behaviour: Only display the file that matches the filename present in the answer of the field. (In the uncompeted journey, there is a "staging" period whereby an applicant may have uploaded a new file but has not re-submitted. In this case the assessor will see 'Not provided' for a period, until resubmission occurs). 

### How to test
_If manual testing is needed, give suggested testing steps_

### DIFF
--- a/pre_award/assess/assessments/models/applicants_response.py
+++ b/pre_award/assess/assessments/models/applicants_response.py
@@ -342,6 +342,7 @@ def _ui_component_from_factory(item: dict, application_id: str):  # noqa: C901
                 quoted=True,
             )
             for key in file_keys
+            if answer == key.split("/")[-1]
         }
         return QuestionAboveHrefAnswerList.from_dict(item, key_to_url_dict)
 


sc15zs:
_In this case the assessor should only see the original file in Assess._

What I see is as soon as a new file is chosen in Apply (and it's not even saved yet), in Assess it no longer shows the file uploaded previously. It says "Not provided". If that's the behaviour we want, then all is good. 

But if we want the assessor to see the original file until the submission is done with the new file, then it's not exactly what we want. 
Again, might be something messed up with my settings, happy to show you tomorrow what I see.

ksp37-dluhc:
> _In this case the assessor should only see the original file in Assess._
> 
> What I see is as soon as a new file is chosen in Apply (and it's not even saved yet), in Assess it no longer shows the file uploaded previously. It says "Not provided". If that's the behaviour we want, then all is good.
> 
> But if we want the assessor to see the original file until the submission is done with the new file, then it's not exactly what we want. Again, might be something messed up with my settings, happy to show you tomorrow what I see.

Hi Zoltan, yep this is the expected behaviour. Sorry in my description I wasn't clear - this is only a partial fix, so although the Assessor will see 'Not provided', hopefully this is only for a brief period and better than the current behaviour where the updated file is shown immediately and could cause confusion. 

There is also the edge case where the applicant re-uploads a new version of the file with the same name. We've discussed some more comprehensive fixes but they may / may not be included in the MVP depending on prioritisation.

sc15zs (approved):
👍

PR #313, opened 2025-03-05
FS-5043 - Adding FAB-specific error content in Authenticator when user is missing required role

wjrm500:
### Ticket

[Appropriate Authenticator content for FAB users when missing role](https://mhclgdigital.atlassian.net/browse/FS-5043)

### Description

We need to add role-based authentication to FAB, as per TDA requirements for FAB moving into production. With the current implementation of the `login_required` decorator from this funding-service-design-utils library, if the user is missing the required role, they get sent to a page that looks like this:

image

As you can see, the content doesn't make sense from the perspective of a FAB user.

To fix, we are going to begin passing a new `source_app` query parameter via the `_failed_role_redirect` function in fsd-utils - see [PR here](https://github.com/communitiesuk/funding-service-design-utils/pull/222). Then, in the Authenticator, we need to respond to this query parameter, conditionally rendering FAB-appropriate content. This PR handles that.

### How to test

Spin up the apps locally (see [Docker Runner](https://github.com/communitiesuk/funding-service-design-docker-runner)) and hit the endpoint `https://authenticator.levellingup.gov.localhost:4004/service/user?roles_required=FSD_ADMIN&source_app=fund-application-builder` - you should see the following content:

image

Meanwhile if you hit the same endpoint but without the `source_app` parameter, or with `source_app` set to something different to `fund-application-builder`, you'll see the original content:

image

### DIFF
--- a/pre_award/authenticator/frontend/templates/authenticator/user/user.html
+++ b/pre_award/authenticator/frontend/templates/authenticator/user/user.html
@@ -3,43 +3,42 @@
 {%- from "govuk_frontend_jinja/components/button/macro.html" import govukButton -%}
 
 {% block content %}
-<div>
-    <h1 class="govuk-heading-xl">Apply for Funding</h1>
-</div>
-<div>
+   {% if source_app == "fund-application-builder" %}
+       <h1 class="govuk-heading-xl">You do not have access to this service</h1>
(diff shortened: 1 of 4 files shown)


nuwan-samarasinghe (approved):
LGTM

PR #328, opened 2025-03-13
fix forms team support url

nuwan-samarasinghe:
fix forms team support url


wjrm500 (approved):
Looks good to me, content as per https://mhclgdigital.atlassian.net/browse/FS-5043 - thought I'd already added this!

wjrm500:
Good point from @samuelhwilliams re this link ideally being a secret. It's not a big deal practically speaking as unauthorised users will just get hit with the page below:

image

But seems like best practice to hide this sort of thing. @nuwan-samarasinghe FYI

PR #340, opened 2025-03-19
Send incomplete applications emails

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net/browse/FLS-1045

Description:
Once a fund round closes, we want to send an email to applicants with unsubmitted applications the questions and their answers.

How to test locally:
1. Create an application (possibly not a CTDF one, because it has only two criteria, but some other which has more)
2. Fill in a few subcriteria, but don't submit it.
3. Change the deadline of the round to a past date (but not more than a month).
4. Open a terminal inside the docker container:   `docker exec -it $(docker ps -qf name="funding-service-pre-award") bash`
5. Run the function:  `invoke apply.send-incomplete-application-emails`
6. You'll probably want to run it more than once, in this case you'll have to delete the event instance from the database, which has been created for the round, when running the function (otherwise it won't send the email the next time you run the function, because there is already an event created for the round, which means that the emails have already been sent.
7. Depending on your local Notify key, the email is either actually sent to the email address, or can be seen on Notify.

Screenshot of the email sent:
Screenshot 2025-04-16 at 10 31 16

Next page:
Screenshot 2025-04-16 at 10 31 30

Confirming the email address:
Screenshot 2025-04-16 at 10 31 41

Next page (lets the user know that the file will be available for 6 months):
Screenshot 2025-04-16 at 10 31 53

Screenshot of the downloaded file:
Screenshot 2025-04-16 at 10 32 44

### DIFF
--- a/apply/tasks.py
+++ b/apply/tasks.py
@@ -7,8 +7,15 @@ from invoke import Context, task  # type: ignore[attr-defined]
 from app import create_app
 from data.crud.accounts import get_account
 from data.crud.applications import get_applications_for_round_by_status
-from data.crud.fund_round_queries import get_rounds_for_application_deadline_reminders, set_application_reminder_sent
+from data.crud.fund_round_queries import (
+    create_event,
+    get_rounds_for_application_deadline_reminders,
+    get_rounds_with_passed_deadline,
+    set_application_reminder_sent,
+)
 from pre_award.application_store.db.models.application.enums import Status
+from pre_award.application_store.db.queries.application.queries import create_qa_base64file, get_application
+from pre_award.fund_store.db.models.event import EventType
 from services.notify import NotificationError, get_notification_service
 
 
@@ -70,3 +77,51 @@ def send_application_deadline_reminders(c: Context) -> None:
     app = create_app()
     with app.app_context():
         send_application_deadline_reminders_impl()
+
+
+def send_incomplete_application_emails_impl() -> None:
+    rounds_with_passed_deadline = get_rounds_with_passed_deadline()
+    incomplete_statuses = [Status.NOT_STARTED, Status.IN_PROGRESS, Status.COMPLETED]
+
+    for round in rounds_with_passed_deadline:
+        applications = get_applications_for_round_by_status(round.id, incomplete_statuses)
+
+        for application in applications:
+            application_with_form_json = get_application(application.id, as_json=True, include_forms=True)
+
+            try:
+                account_id = str(application.account_id)
+                account = get_account(account_id)
+                if not account:
+                    continue
+
+                application_data = create_qa_base64file(application_with_form_json, True)
+
+                notification = get_notification_service().send_unsubmitted_application_email(
+                    email_address=account.email,
+                    application_reference=str(application.reference),
+                    fund_name=round.fund.name_json["en"],
+                    round_name=round.title_json["en"],
+                    application_data=application_data,
+                    contact_help_email=round.contact_email,
+                )
+                current_app.logger.info(
+                    "Sent incomplete application notification %(notification_id)s to %(account_id)s",
+                    dict(notification_id=notification.id, account_id=account_id),
+                )
+            except NotificationError:
+                current_app.logger.exception(
+                    (
+                        "Error sending incomplete application notification to %(account_id)s "
+                        "for fund_id=%(fund_id)s round=%(round_id)s]"
+                    ),
+                    dict(account_id=account_id, fund_id=round.fund_id, round_id=round.id),
+                )
+        create_event(round.id, EventType.SEND_INCOMPLETE_APPLICATIONS, datetime.now(), True)
+
+
+@task
+def send_incomplete_application_emails(c: Context) -> None:
+    app = create_app()
+    with app.app_context():
+        send_incomplete_application_emails_impl()
--- a/data/crud/fund_round_queries.py
+++ b/data/crud/fund_round_queries.py
@@ -1,10 +1,13 @@
+from datetime import datetime, timedelta
 from typing import Sequence
+from uuid import UUID
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import contains_eager
 
 from data.models import Fund, Round
 from pre_award.db import db
+from pre_award.fund_store.db.models.event import Event, EventType
 
 
 def get_fund(fund_short_name: str) -> Fund | None:
@@ -48,4 +51,37 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
 
 def set_application_reminder_sent(round: Round) -> None:
     round.application_reminder_sent = True
+    db.session.commit()
+
+
+def get_rounds_with_passed_deadline() -> Sequence[Round]:
+    """
+    Retrieve rounds that have passed their deadline but have not yet had an event
+    created for sending incomplete applications.
+    """
+    now = datetime.now()
+    one_month_ago = now - timedelta(days=30)
+
+    rounds_without_event = (
+        db.session.query(Round)
+        .filter(
+            Round.deadline < now,
+            Round.deadline >= one_month_ago,
+            Round.id.notin_(
+                select(Event.round_id).filter(
+                    Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS, Event.processed.isnot(None)
+                )
+            ),
+        )
+        .all()
+    )
+
+    return rounds_without_event
+
+
+def create_event(round_id: UUID, event_type: EventType, activation_date: datetime, is_processed: bool) -> None:
+    event = Event(round_id=round_id, type=event_type, activation_date=activation_date)
+    if is_processed:
+        event.processed = datetime.now()  # type: ignore
+    db.session.add(event)
     db.session.commit()
--- a/services/notify/__init__.py
+++ b/services/notify/__init__.py
@@ -486,6 +486,37 @@ class NotificationService:
             },
         )
 
+    def send_unsubmitted_application_email(
+        self,
+        email_address: str,
+        application_reference: str,
+        fund_name: str,
+        round_name: str,
+        contact_help_email: str | None,
+        application_data: dict[str, Any],
+        govuk_notify_reference: str | None = None,
+    ) -> Notification:
+        questions = application_data.get("questions_file")
+
+        return self._send_email(
+            email_address,
+            self.APPLICATION_INCOMPLETE_TEMPLATE_ID,
+            personalisation={
+                "name of fund": fund_name,
+                "round name": round_name,
+                "application reference": application_reference,
+                "question": {
+                    "file": questions,
+                    "filename": None,
+                    "confirm_email_before_download": None,
+                    "retention_period": None,
+                },
+                "contact email": contact_help_email,
+            },
+            govuk_notify_reference=govuk_notify_reference,
+            email_reply_to_id=self.REPLY_TO_EMAILS_WITH_NOTIFY_ID.get(contact_help_email),  # type: ignore[arg-type]
+        )
+
 
 def get_notification_service() -> NotificationService:
     return cast(NotificationService, current_app.extensions["notification_service"])
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -65,6 +65,8 @@ jobs:
             command: job
           - deployment: pre-award-send-application-assessment-report
             command: job
+          - deployment: pre-award-send-incomplete-applications-emails
+            command: job
     permissions:
       id-token: write # This is required for requesting the JWT
       contents: read  # This is required for actions/checkout
--- a/copilot/fsd-pre-award-send-incomplete-applications-emails/manifest.yml
+++ b/copilot/fsd-pre-award-send-incomplete-applications-emails/manifest.yml
@@ -1,0 +1,81 @@
+name: fsd-pre-award-send-incomplete-applications-emails
+type: Scheduled Job
+
+on:
+  schedule: "cron(30 9 * * ? *)"
+retries: 0
+timeout: 10m
+
+# We inject `image.location` in our deployment pipeline
+image: {}
+entrypoint: launcher
+command: invoke apply.send-incomplete-application-emails
+
+cpu: 256
+memory: 512
+platform: linux/x86_64
+
+network:
+  vpc:
+    security_groups:
+      - from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-fsdpreawardstoresclusterSecurityGroup
+
+variables:
+  FLASK_ENV: ${COPILOT_ENVIRONMENT_NAME}
+  SENTRY_DSN: "https://80c7f65b54f0eff535777a66b375adf0@o1432034.ingest.us.sentry.io/4508324370317312"
+  AWS_BUCKET_NAME:
+    from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-FormUploadsBucket
+  ASSESSMENT_FRONTEND_HOST: "https://assess.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  SENTRY_TRACES_SAMPLE_RATE: 0.02
+  API_HOST: "fsd-pre-award.${COPILOT_ENVIRONMENT_NAME}.pre-award.local"
+
+  ALLOW_ASSESSMENT_LOGIN_VIA_MAGIC_LINK: false
+  APPLICANT_FRONTEND_HOST: "https://apply.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk" # TODO: remove me when all frontends combined
+  APPLY_HOST: "apply.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  ASSESS_HOST: "assess.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  AUTH_HOST: "account.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  COOKIE_DOMAIN: ".access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  FORM_DESIGNER_HOST: "https://form-designer.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  FORMS_SERVICE_PUBLIC_HOST: "https://application-questions.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  FUND_APPLICATION_BUILDER_HOST: "https://fund-application-builder.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  POST_AWARD_FRONTEND_HOST: "https://find-monitoring-data.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  POST_AWARD_SUBMIT_HOST: "https://submit-monitoring-data.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk"
+  COPILOT_AWS_BUCKET_NAME:
+    from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-FormUploadsBucket
+  REDIS_INSTANCE_URI:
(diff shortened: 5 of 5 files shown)


data/crud/fund_round_queries.py
```diff
@@ -49,3 +55,71 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+        for round in rounds_with_passed_deadline
+        if not db.session.query(Event)
+        .filter(Event.round_id == round.id, Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS)
+        .first()
+    ]
+
+    return rounds_without_event
+
+
+def get_passed_round_applications(round: Round) -> Sequence[Applications]:
```
gidsg, line 83: Could we have a more accurate function name for this? Something like `get_incomplete_round_applications`?

data/crud/fund_round_queries.py
```diff
@@ -49,3 +55,71 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+    """
```
gidsg, line 84: Could we re-use `get_applications_for_round_by_status()` function for this part?

data/crud/fund_round_queries.py
```diff
@@ -49,3 +55,71 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+
+
+def extract_questions_and_answers(data_list: List[Dict[str, Any]]) -> str:
+    """
+    Function to build a string of questions and answers of application forms
+    """
+    result = []
+    for data in data_list:
+        for question in data["questions"]:
+            if question["category"] == "FabDefault":
```
gidsg, line 106: Why do we treat this `FabDefault` case differently?

data/crud/fund_round_queries.py
```diff
@@ -49,3 +55,71 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+                    if isinstance(answer, str):
+                        answer = strip_html_tags(answer)
+                    result.append(f"{question_text}: {answer}")
+    return ",\n".join(result)
+
+
+def strip_html_tags(text: str) -> str:
+    """Remove HTML tags from a string."""
+    clean = re.compile("<.*?>")
+    return re.sub(clean, "", text)
```
gidsg, line 119: Could we use BeautifulSoup for this? e.g. https://www.crummy.com/software/BeautifulSoup/bs4/doc/#get-text

apply/tasks.py
```diff
@@ -70,3 +79,48 @@ def send_application_deadline_reminders(c: Context) -> None:
...
+                    fund_name=round.fund.name_json["en"],
+                    round_name=round.title_json["en"],
+                    question=summary,
+                    contact_help_email=round.contact_email,
+                )
+                current_app.logger.info(
+                    "Sent incomplete application notification %(notification_id)s to %(account_id)s",
+                    dict(notification_id=notification.id, account_id=account_id),
+                )
+                create_event(round.id, EventType.SEND_INCOMPLETE_APPLICATIONS, datetime.now())
```
ksp37-dluhc, line 111: This isn't actually how the events table was intended to be used; the idea was an (unprocessed) event would be added when the fund was created, and then we would periodically check the events table for those that were past the activation date but unprocessed. Then said event would be updated and marked as processed. 

But what you've done here also works. The only thing that needs to be updated is that the `create_event` method needs a `is_processed` flag that is passed through to the row insertion. And here `is_processed` should be set to `True`.

data/crud/fund_round_queries.py
```diff
@@ -49,3 +55,71 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+    created for sending incomplete applications.
+    """
+    now = datetime.now()
+    one_month_ago = now - timedelta(days=30)
+
+    rounds_with_passed_deadline = (
+        db.session.query(Round).filter(Round.deadline < now, Round.deadline >= one_month_ago).all()
+    )
+
+    rounds_without_event = [
```
ksp37-dluhc, line 72: Just a suggestion, but you could probably do this in a single query. Something like

```
db.session.query(Round)
.join(Event, Event.round_id == round.id, Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS, Event.processed != None, outer=True)
.filter(Round.deadline < now, Round.deadline >= one_month_ago,  Event.id == None).all()
```
I'm pretty sure some of the syntax is incorrect and it'll definitely need some testing!

apply/tasks.py
sc15zs: The Event model doesn't have an `is_processed` field, just a `processed`, which is of a datetime:
`processed = Column("processed", DateTime(), nullable=True)`

Do you want a new is_processed boolean flag added?

ksp37-dluhc: Oops sorry yes, it's a datetime. So the method argument should be `is_processed`, and if set to True, you set the current datetime in the processed field. If False, it's just left as null.

data/crud/fund_round_queries.py
sc15zs: Changed it to use BeautifulSoup.

data/crud/fund_round_queries.py
sc15zs: This function is now deleted, the get_applications_for_round_by_status() function is used to fetch the applications of a round with given statuses.

data/crud/fund_round_queries.py
sc15zs: The only reason for this was to not have the "Do you want to mark this section as complete?" questions and the True / False answers added to the string of questions and answers. 
In the forms all "regular" questions have:   "category": "FabDefault",   the "Do you want to mark this section as complete?" question has: "category": None.

Now I removed this condition, so this question and the answer is also added to the string we use in the email.

apply/tasks.py
sc15zs: Changed it as described.

data/crud/fund_round_queries.py
sc15zs: Sorry Kishan, I wouldn't change it, it's way too complicated for me to follow...
I'd rather have it in two separate steps and understand what it's doing when looking back to it any time later, than make it work with a single query and have no idea what it's doing right now... 
Due to the size of the Round and Event tables, I don't think it would have any performance-issues.
I'll change it to a more elegant single query if you want it though.

albertkol: It should be one query. And a perfect opportunity for a pair programming session. SQL Alchemy has 10000 ways of doing one thing. Not sure if filtering inside the join is the most efficient way to do it, I never did it that way.

But what I suggest is:
- define the Round to Event relationship inside the Round model. You can look inside the `Fund` model to see how to define a one to many relationship
- I would rewrite it to be one query, you can try Kishan's example or there might be few more ways this can be done, but one query function should only be one db query

albertkol (commented):
We need to look into how the confirmation email questions/answers file is created. See if we can re-use that.

data/crud/fund_round_queries.py
```diff
@@ -49,3 +53,55 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+                answer = field["answer"]
+                if isinstance(answer, str):
+                    soup = BeautifulSoup(answer, "html.parser")
+                    answer_text = soup.get_text()
+                elif isinstance(answer, bool):
+                    answer_text = str(answer)
+                else:
+                    answer_text = answer
+                result.append(f"{question_text}\n{answer_text}")
+    return "\n\n".join(result)
```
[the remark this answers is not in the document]
albertkol: I found this function that that email is using.. could we maybe use it?
`q_and_a = extract_questions_and_answers(application_data["forms"], application_data["language"])`

`pre_award/application_store/db/queries/application/queries.py` file`create_qa_base64file` function.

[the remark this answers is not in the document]
sc15zs: We are not creating a file for the user to download, we just send an email, with the questions and answers added to  the email body. The email template is attached to the ticket, it has a single placeholder for this data.
https://mhclgdigital.atlassian.net/browse/FLS-1045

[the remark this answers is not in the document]
sc15zs: _"But I don't really understand what's happening there isn't HTML also a string"_
Yes, it is, the only reason to use BeautifulSoup is to get rid of HTML tags. There are some freetext fields in the form, the text entered here includes HTML tags, which we don't want to display in the email sent to the user.

data/crud/fund_round_queries.py
sc15zs: @albertkol I changed the query, I believe it's hitting the database only once now. 
Could you have a look at it, please?

data/crud/fund_round_queries.py
gidsg: Makes sense, can we check with product what questions we want to include here? (And therefore which version of the logic is preferable?

I am just a bit concerned about the FAB stuff leaking into pre-award implementation, but it's not originating from this change, just first time I noticed it.

data/crud/fund_round_queries.py
```diff
@@ -49,3 +52,30 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+    now = datetime.now()
+    one_month_ago = now - timedelta(days=30)
+
+    subquery = db.session.query(Event.round_id).filter(Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS).subquery()
+
+    rounds_without_event = (
+        db.session.query(Round)
+        .filter(Round.deadline < now, Round.deadline >= one_month_ago, Round.id.notin_(subquery))  # type: ignore
+        .all()
+    )
```
albertkol, line 71: Never used a subquery before. 

I still think a simple join and filter on the joined rows is the right way to go about it, easier to read and reuses logic we used before.

Example from fund_round crud functions:
```python
def get_round(fund_short_name: str, round_short_name: str) -> Round | None:
    round = db.session.scalar(
        select(Round)
        .join(Fund)
        .options(contains_eager(Round.fund)). # don't need this line
        .where(Fund.short_name == fund_short_name)
        .where(Round.short_name == round_short_name)
        .where(Round.is_not_yet_open.is_(False))
    )
    return round
```

Unless you define the relationship between Round and Event, you might need to mention on the join which column from Round matches to with Event column.

pre_award/apply/helpers.py
```diff
@@ -402,3 +404,24 @@ def get_section_feedback_data(application, section_display_config):
...
+                answer = field["answer"]
+                if isinstance(answer, str):
+                    soup = BeautifulSoup(answer, "html.parser")
+                    answer_text = soup.get_text()
+                elif isinstance(answer, bool):
+                    answer_text = str(answer)
+                else:
+                    answer_text = answer
+                result.append(f"{question_text}\n{answer_text}")
+    return "\n\n".join(result)
```
albertkol, line 427: I still think we don't need this function and perhaps can re-use an existing funciton:

> I found this function that that email is using.. could we maybe use it?
> q_and_a = extract_questions_and_answers(application_data["forms"], application_data["language"])
>
> pre_award/application_store/db/queries/application/queries.py filecreate_qa_base64file function.

I was not suggesting to create a file. I was suggesting to try and re-use the function that creates the content for the file we are emailing.

Isn't this more or less what you're doing? Code is here:
https://github.com/communitiesuk/funding-service-design-utils/blob/main/fsd_utils/mapping/application/qa_mapping.py#L19-L85

albertkol (commented):
These are my opinions on this PR. I wanted to write them down, but it takes some investigation to see if or how they could work.

I think the query function can be just one query.
And I think we are duplicating that algorithm to extract the Q&A's.

If others feel different and want to approve the PR, it's up to them.

data/crud/fund_round_queries.py
sc15zs: OK, changed it to not use subquery.

pre_award/apply/helpers.py
sc15zs: Yes, it's more or less what we're doing, the difference is that if I want to use this function, which returns a nested dictionary of questions and answers per form, I'll have to add some other functions to generate a simple string of questions and answers, and remove the form names, which will be longer than this whole function. 
I'll change it if you say it's better to do it this way.

data/crud/fund_round_queries.py
```diff
@@ -49,3 +52,32 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+    """
+    now = datetime.now()
+    one_month_ago = now - timedelta(days=30)
+
+    rounds_without_event = (
+        db.session.query(Round)
+        .filter(
+            Round.deadline < now,
+            Round.deadline >= one_month_ago,
+            Round.id.notin_(select(Event.round_id).filter(Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS)),
```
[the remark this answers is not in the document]
ksp37-dluhc: Sorry I just saw Albert's comment on the same area. My thoughts: the inefficiency was due to several queries, so as long as we use a single query, I don't think it matters too much about the approach of the (single) query itself. The tables aren't big enough to need to worry about that - so I'd favour choosing something simple and understandable. 

If you're interested, this is a good breakdown of the options : https://www.crunchydata.com/blog/rise-of-the-anti-join

apply/tasks.py
ksp37-dluhc: It was intended to be used as a way to schedule jobs that needed to be executed on a certain date, and also keep track of when the jobs had been completed.

However... it's basically only used for incomplete application emails now. We could move the deadline email to Events and drop the extra columns from the table, but it's probably not worth it now that FAB depends on it. (It was also intended to be used to automatically import users from AD ... but that never panned out).

pre_award/apply/helpers.py
sc15zs: @albertkol 
I've changed it as we've discussed: instead of adding the questions and answers to the email body, now there's only a link in the email to a file, which can be downloaded with the questions and answers. 
It uses the existing function and I've deleted my function to build the questions and answers.

copilot/fsd-pre-award-send-incomplete-applications-email/manifest.yml
```diff
@@ -0,0 +1,81 @@
...
+
+on:
+  schedule: "cron(30 9 * * ? *)"
+retries: 0
+timeout: 10m
+
+# We inject `image.location` in our deployment pipeline
+image: {}
+entrypoint: launcher
+command: invoke apply.send-incomplete-application-email
```
albertkol, line 12: There is a typo here: `invoke apply.send-incomplete-application-emails`

apply/tasks.py
```diff
@@ -70,3 +77,49 @@ def send_application_deadline_reminders(c: Context) -> None:
...
+
+            try:
+                account_id = str(application.account_id)
+                account = get_account(account_id)
+                if not account:
+                    continue
+
+                notification = get_notification_service().send_unsubmitted_application_email(
+                    email_address=account.email,
+                    application_reference=str(application.id),
```
albertkol, line 100: Application reference is not the `application.id`. It's the `application.reference`.

services/notify/__init__.py
```diff
@@ -485,6 +487,38 @@ def send_report_email(self, file_buffer: io.BytesIO) -> Notification:
...
+        self,
+        email_address: str,
+        application_reference: str,
+        fund_name: str,
+        round_name: str,
+        application_with_form_json: dict[str, Any],
+        contact_help_email: str | None,
+        govuk_notify_reference: str | None = None,
+    ) -> Notification:
+        application_data = create_qa_base64file(application_with_form_json, True)
```
albertkol, line 500: Please provide this as a parameter to the function. Let's not inject `create_qa_base64file` inside the notify functions.

data/crud/fund_round_queries.py
```diff
@@ -49,3 +52,36 @@ def get_rounds_for_application_deadline_reminders() -> Sequence[Round]:
...
+            Round.deadline < now,
+            Round.deadline >= one_month_ago,
+            Round.id.notin_(
+                select(Event.round_id).filter(
+                    Event.type == EventType.SEND_INCOMPLETE_APPLICATIONS, Event.processed.isnot(None)
+                )
+            ),
+        )
+        .all()
+    )
```
albertkol, line 77: Checked it and it is indeed 1 query.

apply/tasks.py
```diff
@@ -70,3 +77,49 @@ def send_application_deadline_reminders(c: Context) -> None:
...
+                    fund_name=round.fund.name_json["en"],
+                    round_name=round.title_json["en"],
+                    application_with_form_json=application_with_form_json,
+                    contact_help_email=round.contact_email,
+                )
+                current_app.logger.info(
+                    "Sent incomplete application notification %(notification_id)s to %(account_id)s",
+                    dict(notification_id=notification.id, account_id=account_id),
+                )
+                create_event(round.id, EventType.SEND_INCOMPLETE_APPLICATIONS, datetime.now(), True)
```
albertkol, line 110: I suspect this is a logic mistake. You are created an `event` for each application unfinished in the round, instead of just one event per round.

Should probably outside of the `for`.

albertkol (commented):
You can see here in my local `event` table. Multiple events per round.
image

Should be only one event per round, if I understand right how events work.

copilot/fsd-pre-award-send-incomplete-applications-email/manifest.yml
```diff
@@ -0,0 +1,81 @@
...
+  MAINTENANCE_END_TIME: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/FSD_FRONTEND_MAINTENANCE_END_TIME
+  RSA256_PUBLIC_KEY_BASE64: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/RSA256_PUBLIC_KEY_BASE64
+  RSA256_PRIVATE_KEY_BASE64: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/RSA256_PRIVATE_KEY_BASE64
+  GOV_NOTIFY_API_KEY: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/GOV_NOTIFY_API_KEY
+
+environments:
+  prod:
+    variables:
+      FLASK_ENV: production
+      ASSESSMENT_FRONTEND_HOST: "https://assessment.access-funding.levellingup.gov.uk"
```
israr-ulhaq, line 70: Can we change the domain please, i.e., communities ?

sc15zs: Good catch, thanks Israr. 
@gidsg @albertkol @ksp37-dluhc All the other scheduled tasks have the levellingup domain in their manifest, and not the communities one. 
Should I change the domains in this task's manifest file? If yes, all of them?

data/crud/fund_round_queries.py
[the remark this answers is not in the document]
sc15zs: This function is now deleted, and the existing _create_qa_base64file_ function is used to get the application questions and answers.

services/notify/__init__.py
sc15zs: Added _application_data_ as a parameter.

apply/tasks.py
sc15zs: Thanks, Albert. Moved it outside of the inner for loop which loops through all the applications.

copilot/fsd-pre-award-send-incomplete-applications-email/manifest.yml
sc15zs: Typo indeed. Thanks for spotting it. Now corrected.

apply/tasks.py
sc15zs: Changed application ID to reference.

data/crud/fund_round_queries.py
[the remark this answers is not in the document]
sc15zs: Now it's a single query.

copilot/fsd-pre-award-send-incomplete-applications-email/manifest.yml
gidsg: Yes, we should be using the new communities URLs now, please take them from the main manifest.

sc15zs: Changed to use the new communities URLs.

albertkol (approved):
👍 QA'd it and it works: 
- it created one Event per Round
- found the email on notify with the Q&A's
- on second round behaved as expected too

You should run it on dev 1st for extra testing before prod.

israr-ulhaq (approved):
Looks great and works as expected! I'd also suggest testing it on the dev and squashing it before merging. Please!

albertkol (approved):
:+1: Re-approving after changing `deploy.yml`

PR #341, opened 2025-03-19
Add is_resubmission to metadata

albertkol:
Add is_resubmission to metadata.

We need to display a tag on form pages that don't have change requests. 
The absence of change requests is not enough to determine if the form filled in is being submitted for the 1st time ever or not.

So adding this boolean to help with it.


sc15zs (approved):
Checked out the branch and tested.
👍

PR #349, opened 2025-03-26
Update sub-criteria page to latest designs

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-1082

Description:

This UI is for the new design flow for uncompleted funds. It includes a "Review Responses" section, with content that varies depending on the stage of the journey. The "Change Requested" box will remain visible until the subcriteria is scored. Once scored, it will display as the historical view, showing all previous change requests.

Screenshot 2025-03-26 at 13 55 03

Screenshot 2025-03-26 at 13 40 19

Screenshot 2025-03-26 at 13 41 50

Screenshot 2025-03-26 at 15 59 09


pre_award/assess/assessments/routes.py
```diff
@@ -1293,9 +1293,13 @@ def display_sub_criteria(
         "assessment_status": assessment_status,
         "pagination": state.get_pagination_from_sub_criteria_id(sub_criteria_id),
     }
-
+    template_name = (
```
ksp37-dluhc, line 1296: This works, but I think `common_template_config` will need some updating. It's passed to both templates and will contain extra key-values not relevant to competed (e.g. unrequested_changes, change_requests etc). 

I think it might be easier just to have two `render_template` calls, and update `common_template_config` to only contain key-values that are genuinely common to both. Then you can pass the extra uncompeted specific parameters to it's `render_template` call.

pre_award/assess/assessments/templates/assessments/uncompeted_sub_criteria.html
```diff
@@ -0,0 +1,123 @@
...
+{% if migration_banner_enabled %}
+{{ migration_banner() }}
+{% endif %}
+
+{{ sub_criteria_heading(sub_criteria)}}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-third">
+    {{ navbar(application_id, sub_criteria, current_theme.id, is_uncompeted_flow_flag=is_uncompeted_flow(fund)) }}
+  </div>
```
ksp37-dluhc, line 62: It looks like this stuff is common to both templates - so maybe it can be moved out into a common sub_criteria template / macro that's then imported and called here instead.

ksp37-dluhc (commented):
Looks good! I've made a couple suggestions on the two templates - but if it risks taking too long, feel free to create ticket for us to tidy up after delivery.

ksp37-dluhc (approved):
Nice!

PR #353, opened 2025-03-27
Change button text when accepting a change-request

sc15zs:
### Description
Change button text when accepting a response (either after a change-request, or without that).


albertkol (approved):
ayyy

israr-ulhaq (approved):

PR #358, opened 2025-03-28
Hide buttons to change request or approve when change request is active

israr-ulhaq:
Dev Ticket : https://mhclgdigital.atlassian.net/browse/FLS-1115
Design Reference : https://mhclgdigital.atlassian.net/browse/FLS-1133

Hide both the "Approve" and "Raise another change request" buttons when there is an active change request from the assessor.

Previously, the plan was to disable the "Approve" button on the sub_criteria page when an active change request existed.

Screenshot 2025-04-07 at 11 24 02

### DIFF
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -1334,7 +1334,6 @@ def accept_changes(application_id, sub_criteria_id):
     sub_criteria = get_sub_criteria(application_id, sub_criteria_id)
     assessment_status = determine_assessment_status(sub_criteria.workflow_status, state.is_qa_complete)
 
-    is_approval_or_change_request_allowed(state, sub_criteria_id)
     if not is_approval_or_change_request_allowed(state, sub_criteria_id):
         return abort(403)
 
@@ -1386,7 +1385,6 @@ def request_changes(application_id, sub_criteria_id, theme_id):
     field_ids = [question["field_id"] for question in filtered_questions]
     form = build_request_changes_form(field_ids)
 
-    is_approval_or_change_request_allowed(state, sub_criteria_id)
     if not is_approval_or_change_request_allowed(state, sub_criteria_id):
         return abort(403)
 
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -1303,7 +1303,9 @@ def display_sub_criteria(  # noqa: C901
             unrequested_changes=any(theme.get("unrequested_change") for theme in theme_answers_response),
             change_requests=sub_criteria_change_requests,
             has_document_upload=has_document_upload,
-            has_flag_raised=any(flag.latest_status == FlagType.RAISED for flag in sub_criteria_change_requests),
+            has_active_change_request=any(
+                flag.latest_status == FlagType.RAISED for flag in sub_criteria_change_requests
+            ),
             answers_meta=answers_meta,
             questions={question["field_id"]: question["question"] for question in theme_answers_response},
             state=state,
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -1302,6 +1302,7 @@ def display_sub_criteria(  # noqa: C901
             unrequested_changes=any(theme.get("unrequested_change") for theme in theme_answers_response),
             change_requests=sub_criteria_change_requests,
             has_document_upload=has_document_upload,
+            has_flag_raised=any(flag.latest_status == FlagType.RAISED for flag in sub_criteria_change_requests),
             answers_meta=answers_meta,
             questions={question["field_id"]: question["question"] for question in theme_answers_response},
             state=state,
--- a/pre_award/assess/assessments/routes.py
+++ b/pre_award/assess/assessments/routes.py
@@ -64,6 +64,7 @@ from pre_award.assess.assessments.models.location_data import LocationData
 from pre_award.assess.assessments.models.round_summary import create_round_summaries, is_after_today
 from pre_award.assess.assessments.status import (
     all_status_completed,
+    is_approval_or_change_request_allowed,
     update_ar_status_to_completed,
     update_ar_status_to_qa_completed,
 )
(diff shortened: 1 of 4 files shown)


pre_award/assess/templates/assess/macros/application_feedback.html
```diff
@@ -10,15 +10,22 @@
...
+            {% else %}
+                {{ govukButton({
+                    "text": approve_text
+                }) }}
+                {% set request_changes_url = url_for(
+                    'assessment_bp.request_changes',
+                    application_id=application_id,
+                    sub_criteria_id=sub_criteria.id,
+                    theme_id=current_theme_id
+                ) %}
```
sc15zs, line 27: request_changes_url needs to be set outside of the `else` branch, otherwise requesting another change won't work. It won't have the correct url to go to.

pre_award/assess/templates/assess/macros/application_feedback.html
```diff
@@ -10,15 +10,22 @@
...
-            {% set request_changes_url = url_for(
-                'assessment_bp.request_changes',
-                application_id=application_id,
-                sub_criteria_id=sub_criteria.id,
-                theme_id=current_theme_id
-            ) %}
+            {% if has_flag_raised %}
+                {{ govukButton({
+                    "text": approve_text,
+                     "disabled": true,
```
samuelhwilliams, line 16: question: If a user goes into the HTML and enables the button, then clicks it, will our servers reject that request? Ie do we have business logic preventing submission when there are flags, or are we just enforcing that through the untrusted UI?

israr-ulhaq: Yes, if someone pulls off the 'Inspect Element magic trick,' they can enable the button and proceed. This was the approach for the MVP. I hadn’t thought about that—good catch, Sam! Honestly, I think we’re getting rid of this button anyway, but if we still want to keep it disabled, I’ll use business logic to prevent this. Thanks again!

pre_award/assess/templates/assess/macros/application_feedback.html
israr-ulhaq: Thanks for spotting that, Zoltan! I'll fix it.

albertkol:
So we want to hide both buttons? Meaning they cannot ask for a second change request?

israr-ulhaq:
> So we want to hide both buttons? Meaning they cannot ask for a second change request?

Yes, that’s correct this is what the PM/designer decided. The assessor will need to wait for the applicant’s response before they can raise another request at the sub-criteria level.

hamzabinkhalid (approved):
Looks good 👌

albertkol (dismissed):
Please put in a quick check to prevent using the buttons if they are hidden.

And some tests would be nice too.

pre_award/assess/assessments/routes.py
```diff
@@ -1332,6 +1334,10 @@ def accept_changes(application_id, sub_criteria_id):
     sub_criteria = get_sub_criteria(application_id, sub_criteria_id)
     assessment_status = determine_assessment_status(sub_criteria.workflow_status, state.is_qa_complete)
 
+    is_approval_or_change_request_allowed(state, sub_criteria_id)
+    if not is_approval_or_change_request_allowed(state, sub_criteria_id):
+        return abort(403)
```
albertkol, line 1339: Why do you call the function once above the condition? 🤔

tests/pre_award/assess_tests/test_routes.py
```diff
@@ -892,7 +893,7 @@ def test_route_sub_criteria_acceptance_comment(
 
         token = create_valid_token(test_lead_assessor_claims)
         assess_test_client.set_cookie("fsd_user_token", token)
-
+        is_approval_or_change_request_allowed(mock_get_assessor_tasklist_state, sub_criteria_id)
```
albertkol, line 896: Same question.. why do you call the function in test? Am I missing something? 🤔

israr-ulhaq: Opps removed!

pre_award/assess/assessments/routes.py
```diff
@@ -1302,6 +1303,7 @@ def display_sub_criteria(  # noqa: C901
             unrequested_changes=any(theme.get("unrequested_change") for theme in theme_answers_response),
             change_requests=sub_criteria_change_requests,
             has_document_upload=has_document_upload,
+            has_flag_raised=any(flag.latest_status == FlagType.RAISED for flag in sub_criteria_change_requests),
```
ksp37-dluhc, line 1306: Could we change this to `has_active_change_request` ? Just because flag raised is quite a generic term (not all flags are change requests).

tests/pre_award/assess_tests/test_routes.py
```diff
@@ -1611,7 +1611,7 @@ def test_submit_change_request(
...
```
ksp37-dluhc, line 1614: What does this call do?

tests/pre_award/assess_tests/test_routes.py
```diff
@@ -1652,6 +1652,35 @@ def test_submit_change_request(
...
+        assess_test_client.set_cookie("fsd_user_token", token)
+        mock_get_assessor_tasklist_state.return_value["criterias"][0]["sub_criterias"][0]["status"] = "CHANGE_REQUESTED"
+        post_data = {"field_ids": ["JCACTy"], "reason_JCACTy": "testing"}
+
+        response = assess_test_client.post(
+            f"/assess/application_id/{application_id}/sub_criteria_id/{sub_criteria_id}/theme_id/test_theme_id/request_change",
+            data=post_data,
+            follow_redirects=True,
+        )
+        assert b"Access Denied" in response.data
```
ksp37-dluhc, line 1682: Minor point, but I'd check the status_code is 403 rather the specific content in the error page. 

Perhaps there should also be a test for where the status is COMPLETE ?

pre_award/assess/assessments/status.py
```diff
@@ -59,3 +59,12 @@ def update_ar_status_to_qa_completed(application_id, user_id):
             "Could not create qa_complete record for application %(application_id)s",
             dict(application_id=application_id),
         )
+
+
+def is_approval_or_change_request_allowed(state, sub_criteria_id):
+    """Return True only if sub_criteria's status is in  'COMPLETED' or 'CHANGE_REQUESTED'."""
+    for criteria in state.criterias:
+        for sub in criteria.sub_criterias:
+            if sub.id == sub_criteria_id and sub.status in ["COMPLETED", "CHANGE_REQUESTED"]:
```
ksp37-dluhc, line 68: We can use the enums instead of the raw string (pre_award/assessment_store/db/models/assessment_record/enums.py)

ksp37-dluhc (approved):

PR #378, opened 2025-04-04
FLS-1136 Change requested tag is now capitalised instead of uppercase

ksp37-dluhc:
### Change description
As title. Brings the change requested tag into capitalised form (rather than uppercase) which is consistent with the other status tags. 

### How to test
_If manual testing is needed, give suggested testing steps_

Screenshot 2025-04-04 at 15 15 34

### DIFF
--- a/pre_award/apply/static/src/styles/tasklist.css
+++ b/pre_award/apply/static/src/styles/tasklist.css
@@ -177,11 +177,6 @@ ul li.no-bullet-point {
   color: #0b0c0c;
 }
 
-.govuk-tag.change-requested-tag {
-  background-color: #fcd6c3;
-  color: #6e3619;;
-}
-
 .signout {
   padding-left: 0px;
   margin-top: 0px;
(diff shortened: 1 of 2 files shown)


pre_award/apply/static/src/styles/tasklist.css
```diff
@@ -176,6 +177,11 @@ ul li.no-bullet-point {
   color: #0b0c0c;
 }
 
+.govuk-tag.change-requested-tag {
+  background-color: #fcd6c3;
+  color: #6e3619;;
+}
```
albertkol, line 183: You shouldn't need to add custom background-colour/colour and use default orange from govuk design system.

Can't you keep the class `govuk-tag--orange` instead?

ksp37-dluhc: Yeah I can do, and have updated it now. (I was following the pattern for the other tags, where the colours are specified, but as you suggested using the govuk class is better).

albertkol (approved):
👍

PR #383, opened 2025-04-07
Cham 2nd round configuration for assess

sc15zs:
Configuration for CHAM APPLY round - Assessment.


pre_award/assess/config/themes_mapping.py
```diff
@@ -162,6 +162,14 @@ def ordered_themes(fund_round_short_name):
             "partner_organisations",
             "declarations",
         ],
+        "CHAMAPPLY": [
+            "organisation_details",
+            "upload_your_business_plan",
+            "organisation_information",
```
israr-ulhaq, line 168: any reason we're not using the same theme name as "your-skills-and-experience" here instead of "organisation_information"?

sc15zs: Good catch, thank you. I guess I got confused with the form names in CHAM first round, which has the `organisation_details` added. This should be `registration_details`, because the form name in form-builder-adapter, which is missing from here is `cham-apply-registration-details.json`.

sc15zs: Changed it to the correct name.

israr-ulhaq (approved):

PR #398, opened 2025-04-29
FSPT-342 - Support Form Runner basic auth

wjrm500:
### Ticket

[Tear down the UAT (or test?) environment](https://mhclgdigital.atlassian.net/browse/FSPT-342)

### Description

We want to tear down the UAT environment to save costs. Prod FAB currently uses UAT Form Runner for form previews. We are going to re-point Prod FAB at Test Form Runner to remove the UAT dependency from FAB and unblock teardown. We achieved cross-environment (Prod -> UAT) linkage through simply disabling JWT authentication on the specific endpoints FAB requires, for the UAT environment specifically. We will need to disable JWT authentication on those endpoints in the Test environment next, as part of the FAB re-pointing.

This provoked [an idea](https://github.com/communitiesuk/digital-form-builder-adapter/pull/178#discussion_r2063373158) from @samuelhwilliams - why not add basic auth to Form Runner in pre-prod environments to support this JWT auth disabling while maintaining some level of security?  It's a nice compromise, but because Form Runner interacts with not only FAB but also Form Designer and Apply (Pre-Award, this repo), we need to make sure we handle all interactions across services.

This PR specifically ensures that we pass basic auth credentials through to Form Runner in the backend request we make to it, which is to get the rehydration token. Any requests originating from the client's browser (e.g., for actually viewing the forms) will trigger a dialogue box for credential entry.

### What are the next steps?

Before we merge, we need to make sure `FORM_RUNNER_INTERNAL_HOST` has been added to the AWS Parameter Store across all environments.

Longer-term... [Form Runner - Implement secure namespaced cache for Runner preview mode](https://mhclgdigital.atlassian.net/browse/FLS-1196) is a better solution. But this unblocks UAT teardown in the meantime.


nuwan-samarasinghe (approved):
LGTM

PR #413, opened 2025-05-06
Bugfix - Fix failing Apply E2E tests

wjrm500:
### Description

These tests have been failing since [basic auth was added to Form Runner](https://github.com/communitiesuk/digital-form-builder-adapter/pull/178). This is because the `continue_application` function, which redirects users to the Form Runner, is core to Apply.

By switching `FORMS_SERVICE_PUBLIC_HOST` from a Copilot variable to a secret, we can store a basic auth-embedded version of the Form Runner URL in the AWS Parameter Store, meaning we can automate our way through the authentication barrier. As a byproduct, Apply users in pre-prod environments will also be spared the pain of basic auth. As the main point of basic auth is to block web crawlers, this seems okay.

### Linked TODOs

We will need to add a secret `FORM_RUNNER_EXTERNAL_HOST` to all pre-prod environments - Dev, Test and UAT. This URL will be `https://${BASIC_AUTH_USERNAME}:${BASIC_AUTH_PASSWORD}@application-questions.access-funding.${COPILOT_ENVIRONMENT_NAME}.communities.gov.uk`.


nuwan-samarasinghe (approved):
Nice change

samuelhwilliams (approved):

PR #415, opened 2025-05-07
Migrate remnant levellingup->communities

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-350

This excludes domains in our copilot manifests and email addresses, which may need a bit more effort to translate and so I'll do those separately.


gidsg (approved):
Looks good. I think the main service manifest should be simplifiable now, as some of the alias overrides are duplicative.

PR #417, opened 2025-05-08
Address UnboundLocalError on `search_params`

samuelhwilliams:
https://funding-service-design-team-dl.sentry.io/issues/6529261650

If a user tries to hit `/account` page without passing through a fund/round query param, `search_params` has not been defined and we get a 500 error page.

This fixes that by returning a 404 instead.

I don't believe users are ever directed to /account without URL params, so it seems to mostly happen from an edge-case or URL hacking. It's maybe better for them to get a 404 page than a 500 page. And less noisy for us.

Ideally we'd fix this properly by having an /account page that works with no params - but I leave that as an exercise for the future, if we deem it worth doing.


sc15zs (approved):
Indeed much nicer to get a _Page not found_ than a _Sorry, there is a problem with the service_.
👍

PR #425, opened 2025-05-14
Rename levellingup emails to communities

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-350

All emails sent to @levellingup are delivered to @communities (behind the scenes this was always done as a mail alias). We are no longer DLUHC, so let's scrub these remaining levellingup references.

We've updated the email reply-to addresses in Notify already. The reply-to IDs have not changed.


hamzabinkhalid (approved):

samuelhwilliams:
Doing a test release on dev env 👍

samuelhwilliams:
All went well in dev - will merge and sort this all tomorrow morning 👍

PR #449, opened 2025-05-21
Extend assessment deadline for both CHAM rounds

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1302

Description:
The CHAM fund team requested the assessment deadline to be extended for both their rounds. 
See screenshot of the email confirming this in the ticket!!!

### DIFF
--- a/pre_award/fund_store/config/fund_loader_config/FAB/cham_apply.py
+++ b/pre_award/fund_store/config/fund_loader_config/FAB/cham_apply.py
@@ -92,7 +92,7 @@ LOADER_CONFIG = {
         "deadline": "2025-05-23T19:00:00",
         "application_reminder_sent": False,
         "reminder_date": "2025-05-13T08:00:00",
-        "assessment_deadline": "2025-06-01T23:59:00",
+        "assessment_deadline": "2025-07-18T23:59:00",
         "prospectus": "https://www.gov.uk/government/publications/combatting-hate-against-muslims-fund-prospectus",
         "privacy_notice": "https://www.gov.uk/government/publications/combatting-hate-against-muslims-fund-prospectus/combatting-hate-against-muslims-privacy-notice",
         "contact_email": "chambid@communities.gov.uk",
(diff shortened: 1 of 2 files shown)


israr-ulhaq (approved):

pre_award/fund_store/config/fund_loader_config/FAB/cham_apply.py
```diff
@@ -92,7 +92,7 @@
         "deadline": "2025-05-18T22:59:00",
         "application_reminder_sent": False,
         "reminder_date": "2025-05-13T08:00:00",
-        "assessment_deadline": "2025-06-01T23:59:00",
+        "assessment_deadline": "2025-07-18T23:59:00",
```
gidsg, line 95: I guess this is UTC time? Should we be reading the value as UK local time?

sc15zs: So, that's exactly how the issue happened with the CHAM APPLY round closing an hour earlier that it should have. 
Originally, it was set to close an hour earlier, exactly for the same reason why you mentioned. Because BST is one hour ahead of UTC during summer, so I set it to close at 22:59:00 (thinking that we store date-times in UTC).
However, for some reason, the round closed at 22:59 (local time). 

This is the previous PR, when it has been changed to 23:59 from 22:59, to not have the same issue with closing the assessment earlier:
https://github.com/communitiesuk/funding-service-pre-award/pull/448/files

Then, the fund team requested the deadline to be extended until 18 July.
This current PR is for that. 

And this is the original PR, where I thought that it should be stored in UTC, and therefore added as 22:59 and not 23:59, and therefore we've had the problem of closing an hour early:
`"assessment_deadline": "2025-06-01T22:59:00",`
https://github.com/communitiesuk/funding-service-pre-award/pull/379/files#diff-7d4e97822d3277b7c5a0a0d8aa22365082bc4fdbcbaf56cf21e8c77eac3ef06bR95

PR #451, opened 2025-05-23
Open CHAM fund's APPLY round for 65 minutes.

sc15zs:
Description:
CHAM fund's APPLY round needs to be re-opened as:

open:   23 May 2025, at 17:55.      -->  `"opens": "2025-05-23T16:55:00", `
close:   23 May 2025, at 19:00.     -->   `"deadline": "2025-05-23T19:00:00",`

Note that the open date-time is stored in UTC (1 hour ahead of local time), and the deadline is stored in local time. 
This, together with removing the `to_bst` filters, allows opening and closing the round for the requested period, and also to show the correct deadline in the templates. 

The `get_now_UK_time_without_tzinfo()` function is needed to display the "Window closed" banner correctly, after the round deadline is passed.


gidsg (approved):
Looks good, we should have a follow-up task to make the opens and deadline behaviour consistent across the system!

PR #453, opened 2025-05-28
fix CHAM themes_mapping config.

sc15zs:
Description:
Correcting the themes mapping configuration for CHAM APPLY. 
This is needed to make the "View entire application" link show all details of an application.


israr-ulhaq (approved):
Looks fine!

PR #465, opened 2025-06-23
Change round opening to use local time instead of UTC.

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1344

Description:
Round opening date-time had to be stored in UTC to open it in the desired time. Now we're changing the model's functions to use local time instead of UTC, so that the opening time can also be stored in local time, just like the deadline.


israr-ulhaq (approved):
It Works fine locally but good to test in other env as well.

PR #469, opened 2025-06-25
adding GET function to render 'Application Submitted' page

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1336

Description:
So far after submitting an application, we rendered the "Application submitted" template directly after the POST request has been sent. Refreshing this page resulted in the form being submitted again, and this re-submission was recognised as a response to a change-request. 
To stop this happening, after the POST request now we redirect to a new `application_submitted` route. This route uses a GET request and renders the "Application submitted" template, so now refreshing a page will not result in duplicate submission of the application, but just renders the same template.

### DIFF
--- a/pre_award/apply/default/application_routes.py
+++ b/pre_award/apply/default/application_routes.py
@@ -470,38 +470,84 @@ def submit_application():
 
     submitted = format_payload_and_submit_application(application_id)
 
-    with force_locale(application.language):
-        if round_data.is_expression_of_interest:
-            eoi_decision = submitted.get("eoi_decision")
+    # with force_locale(application.language):
+    #     if round_data.is_expression_of_interest:
+    #         eoi_decision = submitted.get("eoi_decision")
+    #         return render_template(
+    #             "apply/eoi_submitted.html",
+    #             eoi_pass=Decision(eoi_decision) in [Decision.PASS, Decision.PASS_WITH_CAVEATS],
+    #             fund_name=fund_data.name,
+    #             round_name=round_data.title,
+    #             fund_short_name=fund_data.short_name,
+
+    return redirect(
+        url_for(
+            "application_routes.application_submitted",
+            application_id=submitted.get("id"),
+            application_language=application.language,
+            reference=submitted.get("reference"),
+            application_reference=application.reference,
+            email=submitted.get("email"),
+            fund_name=fund_data.name,
+            fund_short_name=fund_data.short_name,
+            round_id=round_data.id,
+            fund_type=fund_data.funding_type,
+            round_short_name=round_data.short_name,
+            assessment_start_date=assessment_start_date,
+            is_eoi=round_data.is_expression_of_interest,
+            eoi_decision=submitted.get("eoi_decision"),
+            round_name=round_data.title,
+            round_prospectus=round_data.prospectus,
+        )
+    )
+
+
+@application_bp.route("/application_submitted/<application_id>", methods=["GET"])
+def application_submitted(application_id):
+    # Get query parameters
+    reference = request.args.get("reference")
+    application_language = request.args.get("application_language")
+    application_reference = request.args.get("application_reference")
+    email = request.args.get("email")
+    fund_name = request.args.get("fund_name")
+    fund_short_name = request.args.get("fund_short_name")
+    round_id = request.args.get("round_id")
+    fund_type = request.args.get("fund_type")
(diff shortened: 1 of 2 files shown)


pre_award/apply/default/application_routes.py
```diff
@@ -471,38 +471,84 @@ def submit_application():
 
     submitted = format_payload_and_submit_application(application_id)
 
-    with force_locale(application.language):
-        if round_data.is_expression_of_interest:
-            eoi_decision = submitted.get("eoi_decision")
+    # with force_locale(application.language):
```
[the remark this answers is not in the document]
sc15zs: Yes, sorry, forgot to remove. Now it's gone.

wjrm500 (changes requested):
Almost there I think, couple of things to mention:

1. I've compared the "Application submitted" screen between main branch and this feature branch and there is one difference - on main branch, the "Sign out" link is available in the header; on the feature branch it is not
2. There are a lot of query parameters passed at the moment - we could reduce this to zero by re-implementing the logic of `submit_application` inside the new route function e.g., `get_application_data`, `get_fund_data`, `get_round_data` and `get_assessment_start`. My feeling is the performance impact of a few extra database calls is negligible compared to the benefits of clean URLs and better architecture

sc15zs:
> Getting error when testing an eoi fund, would be helpful to double check this.

Thanks for raising this. Now it's fixed, eoi_decision had to be converted back to integer.

sc15zs:
> Almost there I think, couple of things to mention:
> 
> 1. I've compared the "Application submitted" screen between main branch and this feature branch and there is one difference - on main branch, the "Sign out" link is available in the header; on the feature branch it is not
> 2. There are a lot of query parameters passed at the moment - we could reduce this to zero by re-implementing the logic of `submit_application` inside the new route function e.g., `get_application_data`, `get_fund_data`, `get_round_data` and `get_assessment_start`. My feeling is the performance impact of a few extra database calls is negligible compared to the benefits of clean URLs and better architecture

@wjrm500 Thanks for noticing these.
1. I didn't just want to add the links to the EOI template, I'll leave it to decide for the PM if they want to do it. Created a ticket for this: https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1355
2. Changed it such that we now do everything in the `application_submitted` function, except for actually submitting the application, which remained in the `submit_application` function.

wjrm500 (changes requested):
The `application_submitted ` function scares me a bit now, I must admit!

I can see that we are no longer passing query parameters, but that data that is emitted from `format_payload_and_submit_application` is required for the templates, so to get that we are now using two approaches - session storage and then replication of logic from the application store endpoint (`SubmitApplicationView`) as backup. This manifests in the following lines:

```
submitted = session.pop("submitted_application", None)
```

```
if not submitted:
    account = get_account(application.account_id)
```

```
application_email = submitted.get("email") if submitted else account.email
```

```
if submitted:
    eoi_decision = submitted.get("eoi_decision")
else:
    submit_view = SubmitApplicationView()
    appl = get_application_data(application_id, as_dict=True)
    eoi_decision = submit_view.get_application_eoi_response(appl)["decision"]
```

This feels quite complex, especially that final bit; I think when I mentioned that we could potentially reduce to zero query parameters, I (wrongly) thought that the `application_id` would be sufficient to get all of the stuff we needed for the templates - so my bad!

I _think_ the best approach here is to pass `email`, `reference` and `eoi_decision` as query parameters. That way we're working with a URL of manageable length, we aren't duplicating database calls, and we don't have to worry about pesky sessions. What do you reckon?

pre_award/apply/default/application_routes.py
```diff
@@ -450,6 +450,23 @@ def continue_application(application_id):
 @verify_round_open
 def submit_application():
     application_id = request.form.get("application_id")
+    submitted = format_payload_and_submit_application(application_id)
+    return redirect(
+        url_for(
+            "application_routes.application_submitted",
+            application_id=application_id,
+            email=submitted.get("email"),
```
wjrm500, line 458: Found then tested locally that Flask correctly encodes query parameters including email addresses automatically so this is fine 👍

wjrm500 (approved):
Looks good! Overall a nice improvement to the code

pre_award/apply/default/application_routes.py
sc15zs: Thank you, Will!

PR #470, opened 2025-06-25
enable MultiFieldInput for applicant export

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-1341

This change includes MultiInputFields in the "Applicant Export Information" CSV file, as required by PFN.
To test, please fill out a PFN application, then go to the assessment side and download the "Applicant Export Information" CSV.

For now I've only added 'profile and spend forecast in the assessor_export but we could add more if grant team asks at the time of complete configuration for PFN in assess.

To review the final csv file have a look at the attachment.
Note : PFN is happy with this format

[assessor_export_2025-06-27_09-14-58_en.csv](https://github.com/user-attachments/files/20943542/assessor_export_2025-06-27_09-14-58_en.csv)

### DIFF
--- a/pre_award/assessment_store/db/queries/assessment_records/queries.py
+++ b/pre_award/assessment_store/db/queries/assessment_records/queries.py
@@ -803,8 +803,10 @@ def get_export_data(  # noqa: C901 - historical sadness
                                         applicant_info[title] += f"({child_title}): {child_value}\n"
 
                                 applicant_info[title].strip()
+                                continue
                             else:
-                                applicant_info[title] = answer
+                                applicant_info[title] = str(answer)
+                            applicant_info[title] = answer
             applicant_info = add_missing_elements_with_empty_values(applicant_info, form_fields, language)
         final_list.append(applicant_info)
 


wjrm500 (commented):
Hey @israr-ulhaq, initial thoughts on the code...

- I can see lots of new content in this `assessment_mapping_fund_round.py` but not sure where to validate it against, is it just a case of cross-referencing against the JSON forms for the new PfN round (have they already been committed?)
- Why are we storing the formatted children as a `\n`-separated string just to split it in the new code in `queries.py` - why not store them in a structured way to minimise code complexity down the line?

A thought on the PR itself - best to rebase main not merge main when updating a branch to avoid merge commits. Also, could you provide screenshots in the description please 🙏

Finally, good to highlight PRs ready for review in the #funding-service-live-services Slack channel

Ta

pre_award/assessment_store/db/queries/assessment_records/queries.py
```diff
@@ -792,7 +792,19 @@ def get_export_data(  # noqa: C901 - historical sadness
...
+                                if title not in applicant_info:
+                                    applicant_info[title] = ""
+                                for child in answer:
+                                    for child_key, child_value in child.items():
+                                        child_title = child_map.get(child_key, child_key)
+                                        # Append each child value as a new line in the same column
+                                        applicant_info[title] += f"({child_title}): {child_value}\n"
+
+                                applicant_info[title].strip()
+                            else:
```
sc15zs, line 806: Just a question: When do we expect the `else` branch to be executed? 
I assume it is to cover if `field_type == "MultiInputField"`, but `answer` is not a list. What would it be in this case?

sc15zs: I don't think we need anything extra added, I just don't know when it wouldn't be a list, but something else, and what type would it be in that case (a string, or something else).

israr-ulhaq: Thank You @sc15zs for spotting this!

PR #471, opened 2025-06-25
remove assessment date for DPIF

israr-ulhaq:
DPIF fund team doesn't want to show the assessment dates to the user on the submission page for this round.
Screenshot 2025-06-25 at 16 51 22


wjrm500 (commented):
Have left one comment, which is actually to do with existing code not your code - but as its on the same line, feels like a good chance to change.

❤️  the screenshot in PR description

Only other comment I have is that it is not clear where this work came from, as there is no linked ticket. Obviously I trust that you aren't just inventing requirements, but good to have an audit trail

israr-ulhaq:
> Have left one comment, which is actually to do with existing code not your code - but as its on the same line, feels like a good chance to change.
> 
> ❤️ the screenshot in PR description
> 
> Only other comment I have is that it is not clear where this work came from, as there is no linked ticket. Obviously I trust that you aren't just inventing requirements, but good to have an audit trail

Thanks Will for a beautiful review, after confirming again from Chanel, this change no more needed after the recent content changes. So I'll close this PR now.

PR #491, opened 2025-07-15
Fix the total requested funding calculation for PFN

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1406

Description:
The "Total funding requested" value was not calculated and shown in the banner so far:
image-20250714-090542

This PR addresses this issue, adds required elements to PFN configuration, extends the existing function to do the calculations for PFN applications, and adds unit test to verify that the calculation is correct. 

The banner then shows correctly the "Total funding requested" v
Screenshot 2025-07-15 at 11 48 48
alue:


hamzabinkhalid (approved):
Great implementation and testing 👌, we need to get it over the line for PFN's live date tomorrow, will need to come back and address the sonar issues later.

PR #492, opened 2025-07-15
Fix Flag teams setup and flag history

israr-ulhaq:
### Ticket

https://mhclgdigital.atlassian.net/browse/FLS-1377

### Changes

This PR fixes a couple of bugs around flag allocation in assessment...

Bug #1
- Currently, if a flag is raised but not allocated to a team, then on the assessor tasklist no flag tab will appear under "Flag history"
- This PR fixes that so even for flags without a team we show them under the tab name "Flagged"

Bug #2
- Currently, the radio buttons for team allocation during flag creation do not appear. This seems to be because the OpenAPI spec is no longer used and therefore does not connect the function `get_available_flag_allocations` to a route (regression related to consolidation) 
- This PR fixes that by wrapping that function in a new endpoint, allowing teams to be retrieved per fund and shown during flag creation

### Miscellaneous

### How to test
With Teams Configured:

Pull this PR.
  Use the CTDF fund.
  Raise a flag and check that the team dropdown appears.
  Make sure the flag shows up in the history.
Without Teams Configured:

  Remove the team setup from routes.py.
  Raise a flag and confirm it still appears in the flag history.

Screenshot 2025-07-15 at 12 10 36

### DIFF
--- a/pre_award/fund_store/api/routes.py
+++ b/pre_award/fund_store/api/routes.py
@@ -302,8 +302,17 @@ def set_event_to_processed(event_id):
     abort(404)
 
 
+@fund_store_bp.get("/funds/<fund_id>/rounds/<round_id>/available_flag_allocations")
+def get_available_flag_allocations_endpoint(fund_id, round_id):
+    return jsonify(get_available_flag_allocations(fund_id, round_id))
+
+
 def get_available_flag_allocations(fund_id, round_id):
     # TODO: Currently teams are hardcoded, move it to database implementation
+    from pre_award.assessment_store.config.mappings.assessment_mapping_fund_round import (
+        CTDF_FUND_ID,
+        CTDF_ROUND_1_ID,
+    )
     from pre_award.fund_store.config.fund_loader_config.cof.cof_r2 import (
         COF_ROUND_2_WINDOW_2_ID,
         COF_ROUND_2_WINDOW_3_ID,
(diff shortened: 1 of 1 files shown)


wjrm500 (dismissed):
Discussed on a call with @israr-ulhaq this morn, main points...

- New code makes sense given context of existing code and achieves desired bug fixes, as demonstrated by testing locally on main branch vs feature branch
- Existing designs + code a bit funky e.g., flag history tabs show all flags separately per team, but we already have all flags shown below, so flag history is more "Flags by team", arguably it'd have just made more sense to just have a team filter to filter the existing flag panels by team; in the code we create this team flag stats object with aggregated stats but then don't use these in the frontend. But yeah we can't solve everything in one PR so no issues with this PR
- Consolidation work appears to mean we bypass the OpenAPI spec (`.yml` file), which meant the `get_available_flag_allocations` function (undecorated) became detached from the route and stopped working silently. We should probably just remove this `.yml` file after making sure all functions are properly implemented; or we should keep it and hook it back up properly, either way we need to do _something_

Potential improvements for next time:

- Make it clear that we are doing bug fixes and not implementing new functionality
- If multiple fixes, one PR might still be fine but have a separate commit per bug fix
- For each bug fix / commit, in the description mention what is the bug, why is it happening and how did we fix

pre_award/fund_store/api/routes.py
```diff
@@ -6,6 +6,10 @@
 from fsd_utils.locale_selector.get_lang import get_lang
 
 from data.models import Round
+from pre_award.assessment_store.config.mappings.assessment_mapping_fund_round import (
```
wjrm500, line 9: We should probably move this import into the function like the other similar imports (if you do this just double check it all still works)

israr-ulhaq: Done, Thank You!

PR #498, opened 2025-07-17
Updating GOV.UK rebranding (new logo, color palette and header/footer updates)

nuwan-samarasinghe:
This PR brings our repository in line with the latest upstream changes to implement the new GOV.UK rebranding.

**Summary of Changes:**

- Updated logo to reflect new GOV.UK branding
- Applied colour palette for consistency across UI elements
- Updated header and footer components to match the new GOV.UK design standards

Notes:

- No functional changes; visual/style updates only

PFN page comparison


app.py
```diff
@@ -351,6 +351,10 @@ def inject_global_constants():
 
         return {}
 
+    @flask_app.context_processor
+    def inject_govuk_rebrand():
+        return {"govukRebrand": True}
+
```
gidsg, line 357: Where is this being used? Seems like we hardcode this to True, and don't see any usage of this? So maybe it could be deleted? Sorry if I'm missing something!

nuwan-samarasinghe: I will keep this since sending for every render is bit hard also I have removed some of the unnecessary govukRebrand values

wjrm500 (changes requested):
Blocked as per https://github.com/communitiesuk/digital-form-builder-adapter/pull/198#pullrequestreview-3037572747

PR #500, opened 2025-07-21
Adding table heading elements

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1402

Description:
Adds `<th>` elements to table in assessment dashboard page. 
Please see details and screenshots in comments of the attached JIRA ticket.


sc15zs:
> Hi @sc15zs, only one thing to mention - when `firstCellIsHeader` is true, bold style is automatically applied to text so we no longer need `\<strong\>` tags, which means we no longer need to resort to `html` instead of `text` for the first cell in each row. E.g., we can have:
> 
> ```
> {{ govukTable({
>                 "caption": "",
>                 "firstCellIsHeader": true,
>                 "rows": [
>                     [{'text': 'Applications received'},
>                     {'text': summary.live_round_stats.submitted if summary.live_round_stats.submitted is not none else
>                     'Unavailable'}],
>                     ],
>                 })
>             }}
> ```

Changed the 'html's to 'text's. Thanks, Will.

wjrm500 (approved):

PR #501, opened 2025-07-21
Remove invalid ARIA attributes from elements.

nuwan-samarasinghe:
Remove invalid ARIA attributes from elements.

**Ticket: https://mhclgdigital.atlassian.net/browse/FLS-1399**

No functional change


nuwan-samarasinghe:
@wjrm500 here we should not remove all the places but this does not have a role & not needed to be there for the screen readers

PR #502, opened 2025-07-21
implement comments-export for single application

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1376

Description:
We have previously added the functionality to export all comments of all applications. 
This PR extends this functionality, to allow users to export all comments of a single application.
The link has been added to the application tasklist page on ASSESS:
Screenshot 2025-07-21

Example export file:
[comments_export_CHAM_REG_Application_ID_423f60e9-e606-4edb-aa9f-e5a22a436257_21-07-2025.xlsx](https://github.com/user-attachments/files/21344952/comments_export_CHAM_REG_Application_ID_423f60e9-e606-4edb-aa9f-e5a22a436257_21-07-2025.xlsx)

How to test:
In ASSESS, go to any application and add a few comments. Both application-level comments and sub-criteria comments. 
Click the "Export assessment comments" link, and a file should be generated, including all comments of the given application.

### DIFF
--- a/pre_award/application_store/db/queries/comments/queries.py
+++ b/pre_award/application_store/db/queries/comments/queries.py
@@ -13,12 +13,14 @@ from pre_award.assessment_store.db.models.comment.comments_update import Comment
 from pre_award.db import db
 
 
-def retrieve_all_comments(fund_id, round_id):
+def retrieve_all_comments(fund_id, round_id, application_id=None):
     filters = []
     if fund_id:
         filters.append(Applications.fund_id == fund_id)
     if round_id:
         filters.append(Applications.round_id == round_id)
+    if application_id:
+        filters.append(Applications.id == application_id)
 
     results = (
         db.session.query(CommentsUpdate, Comment, Applications, Account)
@@ -26,7 +28,7 @@ def retrieve_all_comments(fund_id, round_id):
         .join(Applications, Comment.application_id == Applications.id)
         .join(Account, Comment.user_id == cast(Account.id, String))
         .filter(*filters)
-        .order_by(Comment.application_id, Comment.sub_criteria_id)
+        .order_by(Comment.application_id, Comment.sub_criteria_id, Comment.date_created)
         .all()
     )
 
@@ -45,6 +47,13 @@ def retrieve_all_comments(fund_id, round_id):
         sub_criteria = get_sub_criteria(first_application_id, sub_criteria_id)
         sub_criteria_name_map[sub_criteria_id] = sub_criteria.name if sub_criteria.name else None
 
+    comments_list = build_comments_list(results, sub_criteria_name_map)
+
+    return comments_list
+
+
+def build_comments_list(results, sub_criteria_name_map):
+    """Build a list of comment dictionaries from query results."""
     comments_list = []
     for cu, comment, application, commenter_account in results:
         comments_list.append(
@@ -63,7 +72,6 @@ def retrieve_all_comments(fund_id, round_id):
                 "Comment": cu.comment,
             }
         )
-
     return comments_list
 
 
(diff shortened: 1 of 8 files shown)


pre_award/application_store/db/queries/comments/queries.py
```diff
@@ -13,20 +13,22 @@
...
+        filters.append(Applications.id == application_id)
 
     results = (
         db.session.query(CommentsUpdate, Comment, Applications, Account)
         .join(Comment, CommentsUpdate.comment_id == Comment.id)
         .join(Applications, Comment.application_id == Applications.id)
         .join(Account, Comment.user_id == cast(Account.id, String))
         .filter(*filters)
-        .order_by(Comment.application_id, Comment.sub_criteria_id)
+        .order_by(Comment.application_id, Comment.sub_criteria_id, Comment.date_created)
```
wjrm500, line 31: When I test locally, seems like comments aren't actually getting ordered by date_created within their subcriteria, see screenshot:

Image

wjrm500 (changes requested):
Looking really good - works well on test and code all makes sense.

Apart from the potential issue with comments not being ordered correctly by date_created in the export (see my comment), there are a couple of small things to flag which I discovered while testing, which to be honest fall outside the scope of this PR (as you are leveraging existing functionality), but I'll mention them anyway for future reference:

- If you edit a comment, all versions of that comment appear in the export, with no way of identifying which comment is the visible one that appears on the web page
- The field "Sub-criteria ID" in the export feels unnecessary esp. as we already show "Sub-criteria name"
- If an application has no comments, you get a completely blank spreadsheet, without even a header row - I'd have expected a header row even with no data

But in terms of things to action on this PR before I can approve, just the comment ordering thing - and also, please add a test/s. Thanks 🏂

sc15zs:
> > Looking really good - works well on test and code all makes sense.
> > Apart from the potential issue with comments not being ordered correctly by date_created in the export (see my comment), there are a couple of small things to flag which I discovered while testing, which to be honest fall outside the scope of this PR (as you are leveraging existing functionality), but I'll mention them anyway for future reference:
> > 
> > * If you edit a comment, all versions of that comment appear in the export, with no way of identifying which comment is the visible one that appears on the web page
> > * The field "Sub-criteria ID" in the export feels unnecessary esp. as we already show "Sub-criteria name"
> > * If an application has no comments, you get a completely blank spreadsheet, without even a header row - I'd have expected a header row even with no data
> > 
> > But in terms of things to action on this PR specifically before I can approve, just the comment ordering thing - and also, please add a test/s. Thanks 🏂
> 
> @wjrm500 Hi Will, Thanks for the thorough testing. Noticing issue with ordering is a really good catch. Now it should be sorted. The problem was that we have a separate Comment and a CommentsUpdate model. It is the CommentsUpdate which actually holds the _comment text_ added by a user. The returned queryset was sorted by application IDs, sub-criteria IDs, and finally by Comment created_by date. But in the generated file the CommentsUpdate's created_by date was added, which is different from the Comment created by date, if a comment has been updated. Now, the queryset is sorted by the CommentsUpdate's created_by date (after by application ID and sub-criteria ID), which is as it should be.
> 
> Regarding the other points you've raised:
> 
> 1. If you edit a comment, all versions are added to the export: that's the expected behaviour, because on the the webpage also all versions of an edited comment is shown. This is to allow users to see who made a comment and what changes have been made, I guess for audit purposes. It's true for both application-level and sub-criteria comments.
> 2. Removed the sub-criteria ID column, it is indeed unnecessary, when we also have the sub-criteria name in the file.
> 3. You're right about being weird not to have a heading row in the generated file, when there is no data to be added. Now it's changed that the heading row is added in these cases.
> 
> Regarding the unit tests: working on it now.

@wjrm500 
Hi Will,
Now I've added a unit test.

PR #503, opened 2025-07-21
Update reminder date for DPIF R4

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1437

Descripiton:
This PR is only to update reminder date for DPIF R4.
The fund team has chosen to set up only one round, but divide it into two "cohorts". 
The deadline for applications to be considered in the first cohort is 29 July 2025, and a reminder email would be sent out 7 days before this date to applicants (tomorrow). 
We don't want this email to be sent, that's why we update the reminder date to 23 October 2025.

PS:
We don't yet update the application deadline to 30 October 2025, we will do it only on 29 July 2025 (as well as the round "open" date to 29 July 2025.


wjrm500 (approved):
Looks correct but I'm not fully familiar with the process of updating configuration for live rounds, approving but if you wouldn't mind taking the time to briefly explain how you will now propagate this to the database, or whether it just gets auto-loaded and if so how, then that would be appreciated - thanks!

PR #507, opened 2025-07-23
Refactor comment querying

wjrm500:
### Background

While reviewing Zoltan's PR https://github.com/communitiesuk/funding-service-pre-award/pull/502, I noted that the current code architecture around querying comments was wonky. For example...

- Comments are an Assess concern and yet the querying logic was in the application_store module
- Unnecessary usage of the application store's Applications model in a recently-added comment query function (`retrieve_all_comments`) meant further application store tie-in
- The functions `get_comments_from_db` and `retrieve_all_comments` were both doing roughly the same thing, as you can see just from their names

### Description of changes

- **Moving retrieve_all_comments and related functions from application store to assessment store**
- **Use AssessmentRecord not Applications model when retrieving all comments** - retrieving comments happens in the assessment store scope. Using the Applications model from application store breaks that separation. We don't need to do that as we have AssessmentRecord in scope, with application reference, fund and round ID, so let's use that instead. While longer-term consolidation is appropriate, for now Pre-Award code is divided up according to the extinct microservices, and we should follow this pattern here.
- **Refactoring assessment store comment queries module to unify comment getting methods** - get_comments_from_db and retrieve_all_comments were doing a broadly similar thing, just returning the data in a different format depending on whether it is intended for web page display or Excel export. We can therefore have a single query function and then separate format functions that operate on its output. The interface remains unchanged.
- **Renaming comment-getter methods in assessment store for clarity and consistency**

### Notes

I did think about separating out the functions in the new assessment store queries.py into a reduced queries.py and a new services.py, but it was difficult to untangle the logic and my initial implementation didn't seem to achieve much by way of separation, so I decided to leave that for the future.


nuwan-samarasinghe (approved):
Nice change gone through the code with @wjrm500 and also tested it locally 🥇

PR #510, opened 2025-07-28
FLS-1411 - Improve error message clarity

wjrm500:
### 🎫 Ticket

[[Assessment] Improve clarity of error messages for form validation.](https://mhclgdigital.atlassian.net/browse/FLS-1411)

### 🌏 Background

When a user is creating a tag in Assess and does not select any of the radio buttons for the "Tag purpose" field, the default error message "This field is required" is shown. This was flagged in the accessibility audit.

### ♻️ Changes

- The error message "Select a tag purpose" is now used instead
- The other error messages on the "Create tag" page have been improved: "Provide a value for the tag." has been replaced with the more conventional "Enter a tag name", while "Invalid characters in value" has been replaced with the more specific "Tag name can only include letters, numbers, apostrophes, hyphens and spaces"

### DIFF
--- a/pre_award/assess/tagging/forms/tags.py
+++ b/pre_award/assess/tagging/forms/tags.py
@@ -29,7 +29,7 @@ class NewTagForm(FlaskForm):
 
     type = RadioField(
         "type",
-        validators=[InputRequired()],
+        validators=[InputRequired(message="Select a tag purpose")],
     )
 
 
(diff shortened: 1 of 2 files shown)


israr-ulhaq (approved):
'Enter a tag name' error is more readable than 'provide a value'. Nice change 💯

PR #513, opened 2025-08-01
FLS-1441 adding FSD_ADMIN into a DEBUG_USER role

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FLS-1441**

#### How Login Works Normally (SSO Path)

When a user tries to log in:

- The system checks their credentials using Active Directory (a Microsoft directory service that manages users).
- If the user doesn't already exist in the app's database:
  - The system creates the user in the app.
  - It also creates and assigns the user the correct roles (permissions) based on their AD profile.

#### The Problem Happens in “Development Mode”

There’s a special “development mode” (used by developers for testing). In this mode:
- A flag called DEBUG_USER_ON is set to True.
- Instead of reading real roles from the database, the system uses a hardcoded user called DEBUG_USER.
- So, when it creates a login session (technically, a "token" or cookie), it only includes the roles defined in DEBUG_USER.

#### Why This Causes an Error

In the FAB system:
- Some parts of the app require a special role called FSD_ADMIN.
- But the DEBUG_USER didn't have that role listed.
- So, even if everything else worked, the login fails because the app can’t find the required role in the token.
- This leads to an authentication error.

#### What the Fix Was (Pull Request)

To fix this in the code:
- Missing role FSD_ADMIN added into the DEBUG_USER.
- Now, when logging in during development mode, the system includes that role in the token.
- This lets the login succeed and gives the developer access to all the necessary parts of the app.

### How to test

This occurs when accessing FAB via either the [root](https://fund-application-builder.communities.gov.localhost:3011/) or the [dashboard](https://fund-application-builder.communities.gov.localhost:3011/dashboard). use these links to login in local environment and now should be able to login without an issue

### DIFF
--- a/pre_award/config/envs/development.py
+++ b/pre_award/config/envs/development.py
@@ -133,6 +133,7 @@ class DevelopmentConfig(Config):
             "PFN_LEAD_ASSESSOR",
             "PFN_ASSESSOR",
             "PFN_COMMENTER",
+            "FSD_ADMIN",
         ],
         "highest_role_map": {
             "FFW": DEBUG_USER_ROLE,


sc15zs:
Hi Nuwan,
Shouldn't it be added to docker-runner's development.py file?
```
DEBUG_USER = {
        "full_name": "Development User",
        "email": "dev@communities.gov.uk",
        "roles": ["FSD_ADMIN"],
        "highest_role_map": {},
    }
```
I don't know the reason why, but for me only works that way. Even if I do `make pre up`, it won't add this role to DEBUG_USER when I try from pre-award's development.py file.

wjrm500:
Hi @nuwan-samarasinghe @sc15zs is partially right, we will need to also add `FSD_ADMIN` to the debug user config in FAB's development.py. This is because if we access the FAB dashboard directly (at `/dashboard`) in FAB we don't get roles from Pre-Award, instead we take the roles from the FAB debug user config, which currently does not contain `FSD_ADMIN`. This Pre-Award change only works if we access FAB via the root `/`.

wjrm500:
> Why do we have discrepancy in between these endpoints I think thats wrong we have to fix that @wjrm500 there should not be 2 ways of going inside the app both should go with pre-award-auth or they way you mentioned

I think the current situation is okay - the discrepancy you mentioned only happens during local development - and here, when accessing `/` we redirect to `/login` and hit the Authenticator service, thus the need for this Pre-Award PR to return `FSD_ADMIN` in the roles; and when accessing `/dashboard` we bypass Authenticator and should take `FSD_ADMIN` from the FAB debug user config (hence need for FAB PR). This duality - testing Authenticator connection but bypassing if needed - makes some kind of sense to me during local development, so happy to keep as-is.

PR #521, opened 2025-08-12
FLS-1307 Delete application data for unsuccessful COF applicants in accordance with privacy policy

nuwan-samarasinghe:
**Ticket: [Remove the data of unsuccessful applicants from COF](https://mhclgdigital.atlassian.net/browse/FLS-1307)**

### Change description
As MHCLG, we need to remove application data for users who were unsuccessful in their grant applications, in line with our privacy notice.

This action ensures compliance with our data retention responsibilities under GDPR, even though there is no explicit legal requirement to notify data subjects when deletion occurs unless it is in response to a request. Our data protection team (consulted in April) confirmed that data should be removed in a “timely manner,”.

This PR implements backend logic to identify and securely delete data associated with unsuccessful applications.

### How to test
Create an application and submit it once done delete the data using this script it will delete all the user input data

### DIFF
--- a/pre_award/application_store/db/queries/application/queries.py
+++ b/pre_award/application_store/db/queries/application/queries.py
@@ -4,7 +4,7 @@ import string
 from datetime import datetime, timezone
 from io import BytesIO
 from itertools import groupby
-from typing import Optional
+from typing import Dict, Iterable, List, Optional
 
 from flask import current_app
 from fsd_utils import extract_questions_and_answers, generate_text_of_application
@@ -48,10 +48,22 @@ def get_application(app_id, include_forms=False, as_json=False) -> dict | Applic
         return row
 
 
-def get_application_by_reference(app_ref) -> dict | Applications:
-    stmt: Select = select(Applications).filter(Applications.reference == app_ref and Applications.is_deleted == false())
-    stmt.options(noload(Applications.forms))
-    return db.session.scalars(stmt).unique().one()
+def get_applications_by_references(app_refs: Iterable[str]) -> Dict[str, Applications]:
+    app_refs = list(app_refs)
+    if not app_refs:
+        return {}
+
+    stmt = (
+        select(Applications)
+        .where(
+            Applications.reference.in_(app_refs),
+            Applications.is_deleted == false(),
+        )
+        .options(noload(Applications.forms))
+    )
+
+    rows: List[Applications] = list(db.session.execute(stmt).scalars())
+    return {app.reference: app for app in rows}
 
 
 def get_applications(filters=None, include_forms=False, as_json=False) -> list[dict] | list[Applications]:
--- a/pre_award/application_store/db/queries/application/queries.py
+++ b/pre_award/application_store/db/queries/application/queries.py
@@ -8,7 +8,7 @@ from typing import Optional
 
 from flask import current_app
 from fsd_utils import extract_questions_and_answers, generate_text_of_application
-from sqlalchemy import exc, func, select
+from sqlalchemy import exc, false, func, select
 from sqlalchemy.dialects.postgresql import insert as postgres_insert
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload, load_only, noload
@@ -30,7 +30,7 @@ from pre_award.db import db
 
 
 def get_application(app_id, include_forms=False, as_json=False) -> dict | Applications:
-    stmt: Select = select(Applications).filter(Applications.id == app_id and Applications.is_deleted == False)
+    stmt: Select = select(Applications).filter(Applications.id == app_id and Applications.is_deleted == false())
 
     if include_forms:
         stmt.options(joinedload(Applications.forms))
@@ -49,7 +49,7 @@ def get_application(app_id, include_forms=False, as_json=False) -> dict | Applic
 
 
 def get_application_by_reference(app_ref) -> dict | Applications:
-    stmt: Select = select(Applications).filter(Applications.reference == app_ref and Applications.is_deleted == False)
+    stmt: Select = select(Applications).filter(Applications.reference == app_ref and Applications.is_deleted == false())
     stmt.options(noload(Applications.forms))
     return db.session.scalars(stmt).unique().one()
 
@@ -57,7 +57,7 @@ def get_application_by_reference(app_ref) -> dict | Applications:
 def get_applications(filters=None, include_forms=False, as_json=False) -> list[dict] | list[Applications]:
     if filters is None:
         filters = []
-    stmt: Select = select(Applications).filter(Applications.is_deleted == False)
+    stmt: Select = select(Applications).filter(Applications.is_deleted == false())
 
     if len(filters) > 0:
         stmt = stmt.where(*filters)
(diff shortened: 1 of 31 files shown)


pre_award/assess/templates/assess/macros/application_overviews_table_all.html
```diff
@@ -96,7 +96,7 @@ <h2 class="govuk-error-summary__title">
       <td class="govuk-table__cell">{{ overview.short_id[-6:] }}</td>
       <td class="govuk-table__cell"><a class="govuk-link" data-qa="project_name" id="project-name-{{ loop.index }}"
           href="{{ url_for('assessment_bp.application',application_id=overview.application_id) }}">{{
-          overview.project_name }}</a></td>
+              overview.project_name ~ (' [DELETED]' if overview.is_deleted else '') }}</a></td>
```
hamzabinkhalid, line 99: Just a quick note, the '[DELETED]' label next to the project name is currently not styled in red as per our agreed design. Could you please update the styling to reflect red font [as discussed](https://mhclgdigital.atlassian.net/browse/FLS-1307?focusedCommentId=717929)?

hamzabinkhalid (changes requested):

hamzabinkhalid (changes requested):

hamzabinkhalid:
SonarQube flagged some issues, could you please look at them as well

nuwan-samarasinghe:
I added this exception handling from existing way of handling so I ignore those errors @hamzabinkhalid otherwise we have to either change all or go with the flow

pre_award/assess/templates/assess/macros/application_overviews_table_all.html
nuwan-samarasinghe: @wjrm500 and @hamzabinkhalid do you think red colour work with blue background for me it is hard to identify

hamzabinkhalid:
> I added this exception handling from existing way of handling so I ignore those errors @hamzabinkhalid otherwise we have to either change all or go with the flow

> except ClientError as e:
        current_app.logger.error(e)
        raise Exception(e) from e

Since we're already catching the specific ClientError exception, it would be better to simply re-raise it rather than wrapping it in a generic Exception, which can obscure the original error context.

I also noticed that a similar pattern is used in get_file_for_download_from_aws(). For consistency and clarity, I’d recommend updating both methods to follow the same approach, ideally just re-raising the original exception or wrapping it in a more meaningful custom one if needed.

pre_award/assess/templates/assess/macros/application_overviews_table_all.html
nuwan-samarasinghe: no red colour just without project name we have the [DELETED] placeholder

pre_award/assess/templates/assess/macros/comments.html
```diff
@@ -1,10 +1,10 @@
-{% macro comment(comments, application_id, scroll_comments, display_comment_box, display_comment_edit_box, sub_criteria=False, current_theme=False) %}
+{% macro comment(comments, application_id, scroll_comments, display_comment_box, display_comment_edit_box, state, sub_criteria=False, current_theme=False) %}
```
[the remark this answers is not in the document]
nuwan-samarasinghe: Fixed

hamzabinkhalid:
> 10.7% Duplication on New Code

Sonar is flagging two tests as duplicated, which is technically correct, they share nearly all the same setup and assertions, differing only in the route being tested (resolve_flag vs flag).

We could easily refactor this using @pytest.mark.parametrize to cover both routes in a single test.

pre_award/application_store/db/queries/application/queries.py
```diff
@@ -48,10 +48,16 @@ def get_application(app_id, include_forms=False, as_json=False) -> dict | Applic
         return row
 
 
+def get_application_by_reference(app_ref) -> dict | Applications:
```
wjrm500, line 51: Shouldn't this just be `Applications`? Even if we have to coerce the return, at least we'd know what we're getting out of it. The possibility of a generic `dict` isn't very helpful for consumers of this function

pre_award/apply/default/application_routes.py
wjrm500, line 1: Can see the only route we're blocking access to if the application is deleted is the `tasklist` route - do we want to block other routes as well?

pre_award/assess/assessments/templates/assessments/assessor_tasklist.html
```diff
@@ -38,7 +38,7 @@
         state.fund_name,
         state.fund_short_name,
         state.short_id,
-        state.project_name,
+        state.project_name if not state.is_deleted else state.project_name + " [DELETED]",
```
wjrm500, line 41: I thought we were going for just "DELETED PROJECT" or something? As we're going to wipe the project name if the application gets deleted?

pre_award/db/migrations/versions/014_introduse_newcolumn_for_soft.py
wjrm500, line 1: Should be "introduce", please add a space between "new" and "column" as well 🙏

wjrm500 (commented):
Just doing a drive-by review 🚗 and have left couple of comments - just things I spotted as I was scanning down the code. Nothing in-depth from me as a more thorough review would take quite a bit of time, and I understand Hamza is on this already 🤙 

Re the duplicated code in tests, while I do sympathise with the idea of simply disabling SonarQube for test files (the number of times it's been like "HTTP isn't safe!" when I've just wanted to test example URLs), looking at the new code that's been added here, I do think think heeding SonarQube in this case and adding parametrisation (as Hamza suggests) is a very good shout. We are basically doing the same test again and again after all - hitting some endpoint for a deleted application and verifying that we get 403'ed

pre_award/assess/assessments/templates/assessments/assessor_tasklist.html
nuwan-samarasinghe: Fixed

pre_award/apply/default/application_routes.py
nuwan-samarasinghe: If then @wjrm500 we might need to block lot of places E.g. 

1. Sending Email when the application is not submitted if applicable
2. Should not show the questions pages

hamzabinkhalid (approved):
Thanks for the clear implementation and detailed context, @nuwan-samarasinghe.

Backend logic is well-structured.

Great work addressing all of the issues raised. Approved. 🚀

israr-ulhaq (approved):
Hi @nuwan-samarasinghe , great PR to review on Friday! Excellent work - approving this for merge.

The implementation looks solid and handles the privacy requirements well. I like the consistent is_deleted approach across both apply and assess.

As mentioned in my earlier comment about adding a user message for deleted applications - that's just a nice-to-have UX improvement, not a blocker.

Good job on the comprehensive solution.

Approved ✅

pre_award/apply/default/application_routes.py
```diff
@@ -212,7 +212,7 @@ def tasklist(application_id):
         ttl_hash=get_ttl_hash(Config.LRU_CACHE_TIME),
     )
 
-    if application.status == ApplicationStatus.SUBMITTED.name:
+    if application.status == ApplicationStatus.SUBMITTED.name or application.is_deleted:
```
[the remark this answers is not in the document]
nuwan-samarasinghe: yeah we'll make that later @israr-ulhaq

PR #525, opened 2025-08-26
FLS-1448 - Create form_definition table

wjrm500:
### 🎫 Ticket

[Pre-Award - Create form_definition table](https://mhclgdigital.atlassian.net/browse/FLS-1448)

### 🌍 Background

Funding Service is going to start storing forms in a shared location, to make them accessible to Form Designer, Form Runner and FAB, and end the current architectural weirdness. See [this ticket](https://mhclgdigital.atlassian.net/browse/FLS-1447) for more detail. Specifically, we're going to store form JSONs in a new table `form_definition` and then build a new API to mediate access to it.

This PR creates the table. A new SQLAlchemy model was created, and then the migration was auto-generated using the command `uv run flask db migrate -m "Add form_definition table"`. The migration was applied locally to auto-generate the update to `.current-alembic-head` also. The new SQLAlchemy model exists in a new directory called `form_store`, a sister directory to the existing `account_store`, `application_store`, `assessment_store` and `fund_store` directories. This is because forms are a separate concern and deserve their own space. To enable Alembic to recognise the SQLAlchemy models in this new location, we needed to modify the `env.py` file.


hamzabinkhalid (approved):
LGTM, The form_definition model and migration are well-structured and clearly documented. Good use of UUID, JSONB, and timestamp defaults. Everything looks solid. Approved!

PR #526, opened 2025-08-27
FLS-1449 - Implement forms API

wjrm500:
### 🎫 Ticket

[Pre-Award - Implement forms API](https://mhclgdigital.atlassian.net/browse/FLS-1449)

### 🌍 Background

Forms are going to be stored in a shared location, a new table `form_definition` in the Pre-Award database. We need an API to mediate access to the contents of the table, which will be used by three different services - FAB, Form Designer and Form Runner.

This PR creates that API, including route functions, query functions (data layer) and tests. The new API endpoints are as follows:

- `GET /forms` - Returns a list of all forms as objects including `name`, `draft_json` and `published_json`
- `GET /forms/{name}/draft` - Returns the `draft_json` object
- `GET /forms/{name}/published` - Returns the `published_json` object
- `GET /forms/{name}/hash` - Returns just the hash of the `published_json` for cache validation in Form Runner
- `POST /forms` - Accepts JSON payload containing `name` and `form_json`; creates or updates a form, propagating `form_json` to the `draft_json` field of `form_definition`, and leaving `published_json` empty
- `PUT /forms/{name}/publish` - No request body; publishes a form (copies JSON from `draft_json` column to `published_json`)

### 🚧 Testing

- Tests have been added for both the route functions and the underlying query functions
- The route function tests only need to assert against the API responses rather than against the database, keeping them tightly scoped. This is possible thanks to the fact that the API responds with the output of the `as_dict` method called on the `form_definition` object being operated on
- The query tests assert against the database to give us confidence at the data layer

### DIFF
--- a/pre_award/form_store/api/routes.py
+++ b/pre_award/form_store/api/routes.py
@@ -51,6 +51,10 @@ class FormsView(MethodView):
             name = data["name"]
             form_json = data["form_json"]
 
+            # Validate that name is not empty or whitespace-only
+            if not name or not name.strip():
+                return {"error": "Form name cannot be empty"}, 400
+
             # Validate that form_json is valid JSON if it's a string
             if isinstance(form_json, str):
                 try:
--- a/pre_award/form_store/api/routes.py
+++ b/pre_award/form_store/api/routes.py
@@ -1,5 +1,6 @@
 import hashlib
 import json
+import re
 
 from flask import current_app, request
 from flask.views import MethodView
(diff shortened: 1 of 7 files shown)


pre_award/form_store/api/routes.py
```diff
@@ -0,0 +1,148 @@
...
+                return {"error": "Missing required fields: name, form_json"}, 400
+
+            data = request.get_json()
+
+            if not data or "name" not in data or "form_json" not in data:
+                return {"error": "Missing required fields: name, form_json"}, 400
+
+            name = data["name"]
+            form_json = data["form_json"]
+
```
israr-ulhaq, line 53: I think we should not allow empty form names, otherwise the API would fail at /publish, /draft, etc. endpoints due to URL routing issues (e.g., /forms//publish returns 404).

worth adding basic validation?

israr-ulhaq (commented):
Great work! Excellent test coverage.

One small suggestion: add validation for empty form names to prevent URL routing issues with endpoints like /forms//publish.

Otherwise, really solid implementation! ✅

israr-ulhaq (approved):
Looks good to me! Code changes are clear and consistent , tested each endpoint and it worked well 🚀

PR #533, opened 2025-09-16
FLS-1450 - Remove JSON from GET /forms endpoint

wjrm500:
### 🎫 Ticket

[Form Designer - Use Pre-Award API for retrieving and publishing forms](https://mhclgdigital.atlassian.net/browse/FLS-1450)

### ♻️ Changes

Previously we returned a list of forms along with all of their attributes from the `GET /forms` endpoint, including `draft_json` and `published_json`, which can be very large. The endpoint is only hit by the Form Designer and the `draft_json` and `published_json` attributes aren't accessed on the requester side, plus the amount of data being sent over the network meant the API call took an age to complete. This change removes those attributes from the response.

What's the link to the ticket? The performance implications of sending everything back in the response were only discovered during implementation / review of the Form Designer changes.

### 🚧 Testing

A test already existed for this endpoint; it has been updated in this PR.

### DIFF
--- a/pre_award/form_store/api/routes.py
+++ b/pre_award/form_store/api/routes.py
@@ -38,7 +38,7 @@ class FormsView(MethodView):
         """GET /forms - Returns a list of all forms"""
         try:
             forms = get_all_forms()
-            return [form.as_dict() for form in forms], 200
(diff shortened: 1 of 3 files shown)


israr-ulhaq (approved):
Good change much lighter responses now.

PR #534, opened 2025-09-16
FLS-1450 - Remove login_required decorator from forms API endpoints

wjrm500:
### 🎫 Ticket

[Form Designer - Use Pre-Award API for retrieving and publishing forms](https://mhclgdigital.atlassian.net/browse/FLS-1450)

### ♻️ Changes

We do not need to authenticate any of the form store API endpoints as they are only exposed internally within the AWS VPC. This follows the pattern of other Pre-Award APIs including application store.


israr-ulhaq (approved):
👍

PR #541, opened 2025-09-26
Add page title to 'Contact us' page

sc15zs:
**Ticket:**
https://mhclgdigital.atlassian.net/browse/FLS-1516

**Description:**
I’ve refactored the _app.py_ file to provide both `get_service_title` (for the core service name) and `get_page_title` (which appends " - GOV.UK") as template functions. This makes the template logic cleaner: we can still use `get_service_title` wherever needed, and we can avoid manual string concatenation to add “GOV.UK” in templates by using the new `get_page_title` function.

This PR only adds a page title to the "Contact us" page.
For all other pages, I will create a separate ticket and liaise with the content designers about the required page titles.

**Screenshot:**
Screenshot 2025-09-26 at 13 59 55

### DIFF
--- a/app.py
+++ b/app.py
@@ -354,42 +354,55 @@ def create_app() -> Flask:  # noqa: C901
 
         return {}
 
+    def _get_service_title():
+        fund, round = None, None
+        if request.view_args or request.args or request.form:
+            try:
+                fund = find_fund_in_request()
+                round = find_round_in_request()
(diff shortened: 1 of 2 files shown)


wjrm500 (approved):
Hi @sc15zs thanks for reducing the scope of the changes to the "Contact us" page only, this implementation is neat

PR #644, opened 2026-03-31
Validate SSO return_path parameter

wjrm500:
### Summary

- Add `_is_safe_return_path` validation to `SsoBase` that rejects paths containing `@`, `://`, or protocol-relative `//` prefixes
- Apply validation in `SsoLoginView.get` before storing `return_path` in the session - unsafe values are discarded as `None`, which the downstream `SsoGetTokenView` already handles gracefully via its `if return_path :=` check
- Add parameterized tests covering malicious paths (`@evil.com`, `//evil.com`, `https://evil.com`) and a valid path (`/dashboard/overview`)
- Fixes an open redirect where a crafted `return_path=@evil.com` would produce a redirect URL with `evil.com` as the hostname due to URL userinfo parsing
- In practice modern browsers block or warn on the `@` trick in HTTPS URLs and the session cookie wouldn't follow the redirect, so real-world exploitability is low - but this is a straightforward defence-in-depth fix

### DIFF
--- a/pre_award/authenticator/api/sso/routes.py
+++ b/pre_award/authenticator/api/sso/routes.py
@@ -41,6 +41,10 @@ class SsoBase:
         parsed_uri = urlparse(url)
         return f"{parsed_uri.scheme}://{parsed_uri.netloc}"
 
+    @staticmethod
+    def _is_safe_return_path(path: str) -> bool:
+        return path.startswith("/") and not path.startswith("//") and "://" not in path and "@" not in path
+
     def build_auth_code_flow(self, authority=None, scopes=None):
         return self._build_msal_app(authority=authority).initiate_auth_code_flow(
             scopes or [], redirect_uri=Config.AZURE_AD_REDIRECT_URI
@@ -95,7 +99,8 @@ class SsoLoginView(SsoBase, MethodView):
 
         if return_app := request.args.get("return_app"):
             session["return_app"] = return_app
-            session["return_path"] = request.args.get("return_path")
+            return_path = request.args.get("return_path")
+            session["return_path"] = return_path if return_path and self._is_safe_return_path(return_path) else None
             current_app.logger.debug(
                 "Setting return app to %(return_app)s for this session", dict(return_app=return_app)
             )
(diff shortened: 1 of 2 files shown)


pre_award/authenticator/api/sso/routes.py
```diff
@@ -41,6 +41,10 @@ def _get_origin_from_url(url):
         parsed_uri = urlparse(url)
         return f"{parsed_uri.scheme}://{parsed_uri.netloc}"
 
+    @staticmethod
+    def _is_safe_return_path(path: str) -> bool:
+        return path.startswith("/") and not path.startswith("//") and "://" not in path and "@" not in path
```
samuelhwilliams, line 46: comment: My LLMs suggest a slightly different implementation using urlparse - for consideration?

```
def is_safe_url(target):
    ref_url = urlparse(request.host_url)
    test_url = urlparse(urljoin(request.host_url, target))
    return test_url.scheme in ('http', 'https') and ref_url.netloc == test_url.netloc
```

am told this is a slightly more 'standard' / 'reliable' approach than simple string matching

wjrm500: my LLM has responded with this:
```
It's a reasonable suggestion but it doesn't actually work here. The function validates return_path in SsoLoginView.get, but the path gets concatenated with a different origin later in SsoGetTokenView.get (the safe_app.login_url origin, not request.host_url). So checking against request.host_url at validation time would be comparing against the wrong domain.

The urljoin approach is also subtler than it looks — urljoin("https://legit.com/", "@evil.com") produces https://legit.com/@evil.com (a path), not a redirect to evil.com. So it would pass the netloc check but the actual vulnerability (string concatenation at line 169) would still produce https://legit.com@evil.com which is malicious. The validation and the usage site use different URL construction methods, so the urljoin-based check doesn't model the actual threat.

The simple string matching directly addresses what the concatenation at line 169 can produce. It's easy to audit and doesn't have edge-case mismatches between how the validator constructs URLs and how the real code constructs them.

I'd stick with what you have.
```

🤖 🔥 wars begin

wjrm500: @samuelhwilliams  
image

PR #651, opened 2026-04-14
Let any assess role satisfy the assess blueprint entry gate

wjrm500:
## What's the problem?

The assessment service sits behind a gate that redirects anyone without a "Commenter" role on a fund away from every assess page, including the landing dashboard. So a user provisioned in Azure AD as a Lead Assessor — and nothing else — couldn't even see the dashboard, despite Lead Assessor being a strictly more privileged role than Commenter. An internal user hit exactly this and had to be granted a Commenter role as a workaround.

The reason it hadn't bitten before is operational: in practice, users are usually added to multiple AD groups (or groups are nested) so that a Lead Assessor also ends up with a Commenter role somewhere. The moment that assumption doesn't hold — a new fund, a one-off provisioning, a different AD tenant configuration — people get locked out of a service they have permission to use.

## What does this PR do?

It relaxes the gate. Instead of requiring a Commenter role specifically, the gate now lets in anyone who holds any recognised assess role (Commenter, Assessor, or Lead Assessor) on any fund. Users with no assess roles at all are still redirected to the "you don't have access" page exactly as before.

## How the assess permission model is layered

There are two layers, and this PR only touches the first:

1. **Entry gate** (`auth_protect`, called from the assess blueprint's `before_request` and from the 404 handler for the assess host — just these two call sites). Binary "do you have any assess access at all?" check.
2. **Per-route decorators** (`check_access_application_id`, `check_access_fund_short_name_round_sn`, `check_access_fund_id_round_id` in `validation.py`). These resolve the fund from the URL, check the user has access to that specific fund via `has_access_to_fund` (which already accepts any of the three roles), and then delegate to `fsd_utils.login_required(roles_required=...)` to enforce the specific role needed for the action — e.g. scoring requires Lead Assessor or Assessor, flag resolution requires Lead Assessor, tagging requires Assessor.

`auth_protect` has no business knowing about specific roles — that's what the per-route decorators are for. This PR makes the gate reflect that split.

## Why is this safe?

The per-route decorators are unchanged. A Lead Assessor who now gets through the gate will be accepted by `has_access_to_fund` on exactly the same basis a Commenter would have been, and the action-specific role checks (Lead Assessor to resolve a flag, Assessor to tag, etc.) run as normal. Anyone who lacks the right role for a specific action will still get a 403 at the route itself.

I also checked every production code path for Commenter-specific logic. The only reference outside the gate is in `round_summary.py`, and it goes the other way — it hides pre-assessment rounds from Commenters, which Lead Assessors pass through without issue. The `is_commenter` property on `AssessmentAccessController` exists but is never referenced.

## Why is this a good idea?

- Fixes a real user-facing bug where people with legitimately higher permissions were being denied access.
- Removes a latent footgun: any future AD provisioning that grants Lead Assessor without also bundling Commenter will no longer silently lock the user out.
- Aligns the gate with `has_access_to_fund`, so there's one shared definition of "has assess access" rather than two subtly different ones.
- Doesn't expand anyone's permissions. Every existing user sees exactly what they saw before; only users who were being incorrectly bounced at the front door are affected.

## Test plan

- [ ] Automated tests pass, including a new case covering a user with only a Lead Assessor role
- [ ] In a deployed environment, a user with only `{fund}_LEAD_ASSESSOR` (no Commenter anywhere) can reach the dashboard
- [ ] Commenter-only and Assessor-only users still reach the dashboard unchanged
- [ ] A user with no assess roles is still redirected to the "no access" page
- [ ] Spot-check an action gated by a specific role (e.g. flag resolution, which requires Lead Assessor) still denies users who lack that role

### DIFF
--- a/pre_award/assess/authentication/auth.py
+++ b/pre_award/assess/authentication/auth.py
@@ -1,40 +1,40 @@
 from flask import g, redirect, request
 from fsd_utils.authentication.models import User
 
+from pre_award.assess.authentication.validation import ASSESS_ROLES
 from pre_award.assess.services.data_services import get_funds
 from pre_award.assess.shared.helpers import get_ttl_hash
 from pre_award.config import Config
 
 
-def auth_protect(minimum_roles_required: list, unprotected_routes: list):
+def auth_protect(unprotected_routes: list):
     """
-    Checks the authentication and authorisation attributes of the
-    user accessing the service and allows them to access the requested
-    resource if authorised, or redirect them to login (or a permissions
-    error message) if not authenticated/authorised.
+    Entry gate for the assess blueprint.
 
-    The function also incorporates support for bypassing authorisation
-    via setting of a DEBUG_USER_ROLE env var which will mimic the effect
-    of a user with that role (only used when FLASK_ENV is "development").
+    A user is granted access if they hold any role recognised by the assess
+    module (see ASSESS_ROLES) on any fund. Fine-grained role requirements for
+    specific actions are enforced by the per-route decorators in
+    ``pre_award.assess.authentication.validation``; the gate's only job is to
+    redirect users with zero assess access away from the blueprint.
+
+    Supports bypassing authorisation via ``DEBUG_USER_ON`` in development.
 
     Args:
-        minimum_roles_required: List[str]
-            - a list of minimum roles a user must at least one of to be authorised
         unprotected_routes: List[str]
-            - a list of routes e.g. ["/"]
-                that can be accessed without authentication
+            - a list of routes e.g. ["/"] that can be accessed without
+              authentication.
 
     Returns:
-        redirect (302) or None if authorised
-
+        redirect (302) or None if authorised.
     """
-    # expand roles to include all fund short names as a prefix, e.g. "COMMENTER" becomes "COF_COMMENTER"
-    minimum_roles_required = [
+    # Every {fund}_{role} string that counts as "has assess access", matching
+    # how tokens are emitted at sign-in (upper-cased, fund-prefixed).
+    permitted_roles = [
         f"{fund.short_name}_{role}".upper()
         for fund in get_funds(
             get_ttl_hash(seconds=Config.LRU_CACHE_TIME)
         )  # expensive call, so cache it & refresh every 5 minutes
-        for role in minimum_roles_required
+        for role in ASSESS_ROLES
     ]
 
     if not g.is_authenticated and Config.FLASK_ENV == "development" and Config.DEBUG_USER_ON:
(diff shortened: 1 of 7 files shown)


wjrm500:
Feels slightly over-engineered (took me some time to mentally parse the logic) but it's... basically fine?

pre_award/assess/authentication/auth.py
```diff
@@ -5,6 +5,17 @@
 from pre_award.assess.shared.helpers import get_ttl_hash
 from pre_award.config import Config
 
+# Role hierarchy from highest to lowest. A user holding a higher role implicitly
+# satisfies a lower minimum (e.g. LEAD_ASSESSOR satisfies a COMMENTER minimum).
+_ROLE_HIERARCHY = ["LEAD_ASSESSOR", "ASSESSOR", "COMMENTER"]
```
sfount: Nothing else in this lib refers to or requires knowledge or the actual roles, a cursory look shows that roles are scattered throughout the code and database as free strings and I'm not sure theres a sensible place for this logic but this is another mapping to follow

sfount: It's unlikely important for this code bases context but if other roles were introduced this logic may not always map. I think "COMMENTER" as "Should have access" is where calling it a hierarchy feels shaky

sfount (commented):
I don't have confidence in the coverage or usage of roles in the code base - I've not tried any of it but a hypothesis is that this change might allow users in (this seems to be hooked up to a before request) and then they might functionally be stopped from doing a lot of actions or seeing pages without the right permissions mapped in other "authentication" places that also check roles.

Although quite strange the documentation does include a steer that users added will need the "COMMENTER" role in order to access the system https://github.com/communitiesuk/funding-service-pre-award/blob/62bd1267fc5156d42a528d7bdf98c2bb9e5258b1/pre_award/frontend/README.md?plain=1#L178-L181 - that might want to be updated if this change does work and users without the "COMMENTER" role are able to carry out the expected functionality.

I would normally be very hesitant to merge this without some reassurance that users set up with the subset of roles are able to perform their expected actions elsewhere in the codebase but with the context of this service I'm happy to approve if we think its sensible after a quick sense check

wjrm500:
@sfount I've completely redone the implementation, as I think we're both agreed the original was a bit naff! The PR description is worth a read as it lays out the change nicely.

sfount (approved):
Nice this iteration keeps the roles defined centrally and makes sure the behaviour will be applied consistently be removing the (previously non-variable) param. I haven't pulled or tried this behaviour locally but the test case proves the goal of the PR and analysis has been done on any other questions raised

PR #653, opened 2026-04-16
Retry connection errors in Assess get_data to fix intermittent Sentry noise on fund dashboard

wjrm500:
### Change description

The Assess fund dashboard occasionally logs `RemoteDisconnected` / `ConnectionError` to Sentry (83 events / 17 users over the last 16 days) when loading data from internal endpoints like `/account/accounts/fund/<code>`. See https://communitiesuk.sentry.io/issues/108884229/?alert_rule_id=463249&alert_type=issue&notification_uuid=6e4f1083-6c33-48d9-ba36-4d1ac840aa68&project=4511038204674128&referrer=slack

The cause is a keep-alive mismatch on the internal self-call hop: Envoy (Service Connect) pools upstream connections for up to an hour, but gunicorn's keep-alive is 2 seconds, so Envoy occasionally reuses a socket that gunicorn has already closed. The request never reaches the server (confirmed - no matching access log for failing trace IDs), and `get_data` swallows the error and returns `None`, leaving the dashboard rendered with blank "Assigned to" data. Users don't see an error page, just missing data that a refresh fixes.

This adds a small retry loop (3 attempts) around the `requests.get` calls in `get_data`, catching `ConnectionError` only. GETs are idempotent so this is safe, and retrying picks up a fresh connection from the pool. Deliberately kept minimal given the service is being sunsetted.

### DIFF
--- a/pre_award/assess/services/data_services.py
+++ b/pre_award/assess/services/data_services.py
@@ -31,32 +31,33 @@ from pre_award.db import db
 from services.notify import get_notification_service
 
 
-def get_data(endpoint: str, payload: Dict = None):
+def _get_with_retry(endpoint: str, payload: Dict = None):
     # Retry transient connection errors (e.g. RemoteDisconnected from the stale
     # upstream sockets that Envoy occasionally reuses between Service Connect and
     # gunicorn's 2s keepalive). GET is idempotent so retrying is safe.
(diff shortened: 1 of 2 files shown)


srh-sloan (approved):
Looks good, nicely tested 👍

PR #658, opened 2026-05-06
Bug fix: show who created the change request, not the last user to interact with it

samuelhwilliams:
### Change description
If the change request was handled by an applicant, we would incorrectly show their name as the person who requested the change. It should always be the name of the user who actually raised the change request.

### How to test
Change the macro calls back to referencing `latest_user_id` and run the tests - they'll fail.

#### Before
image

#### After

image


MarcUsher (approved):
Goooood fix, it's really gnarly combing through code we've not looked at in months/ever 🙌

PR #663, opened 2026-08-13
FLS-1710: Update PII data deletion script

MarcUsher:
https://mhclgdigital.atlassian.net/browse/FSPT-1530

### Change description
As part of our responsibility to safely process and delete data from past grants, this data deletion script was created so we could safely delete application and assessment. From what I understand it wasn't fully tested or ever run in prod, which limits our ability to safely delete data (inc. PII data) from applications and assessments.

In order to make this fit for the current use case, this adds a number of additional checks, logging and fixes up some bugs:
1. Extend the script to allow exclusion of specific applications via the application.id (this is an explicit need for an upcoming grant data deletion where one or more applications/assessments need to be retained while the rest can be safely deleted)
2. Extend the PiiDeletionLog table to keep a record of which applications have been cleared out, which have been retained, and which have failed to be deleted, so that we have a single place to look for an audit trail for data deletion and a complete snapshot of what was cleared and saved.
3. Fix a bug in the S3 file listing to properly paginate the results and not silently miss files if there were folders with >1000 files
4. Fix a timezone bug in order to correctly do date comparisons for round closing/retention periods
5. Catch a few more parts of the application which could contain PII (eg. surveys and feedback forms)
6. Improved logging during the script to make it clear what's being deleted and what's being retained, along with explicit logging of what's being done at each step for each deletion.

- [X] Unit tests and other appropriate tests added or updated
- ~[ ] README and other documentation has been updated / added (if needed)~
- [X] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
- Set up some applications & assessments for a test fund (eg. Crash Test Dummy Fund CTDF) in various states (submitted and unsubmitted), and run the script locally with some application IDs to be excluded and observe that the correct applications and assessment records are deleted/scrubbed, and the correct applications and assessments are retained.

### DIFF
--- a/common/utils/date_time_utils.py
+++ b/common/utils/date_time_utils.py
@@ -15,4 +15,4 @@ def get_now_UK_time_without_tzinfo() -> datetime:
     Returns the current date and time, using London time, but without timezone info so it can be compared
     to a date from the db.
     """
-    return get_now_from_utc_time_without_tzinfo().astimezone(ZoneInfo("Europe/London")).replace(tzinfo=None)
+    return datetime.now(ZoneInfo("Europe/London")).replace(tzinfo=None)
--- a/data/models.py
+++ b/data/models.py
@@ -7,7 +7,7 @@ from fsd_utils.simple_utils.date_utils import (
     current_datetime_after_given_iso_string,
     current_datetime_before_given_iso_string,
 )
-from sqlalchemy import ColumnElement, Enum, ForeignKey, UniqueConstraint, func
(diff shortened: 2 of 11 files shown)


samuelhwilliams (approved):
Nice work on this - bit of a beast of a codebase to dive into since so long away from it. The dry run can give us some good additional confidence - v sensible to have improved its output.

PR #664, opened 2026-08-20
FLS-1710: Ignore retention on PII deletion script

MarcUsher:
Ticket: https://mhclgdigital.atlassian.net/browse/FSPT-1530

This adds a boolean flag to the PII deletion script to override and ignore the retention period for the submitted and unsubmitted applications.

This is needed because some privacy policies specify dates for successful and unsuccessful applications rather than submitted/unsubmitted, but we have no concept of successful/unsuccessful in the service or the codebase but might have a need to delete them according to these definitions.

The flag set to True overrides both retention period checks and allows deletion of all submitted and unsubmitted applications.


samuelhwilliams (approved):
We paired on this and tested it locally - will let us deal with CHAM's retention policy that differentiates between successful and unsuccessful applications

